PR #9, opened 2017-05-23
JRVLT-177 add test

kwin:
simple file aggregate not correctly overwriting children


kwin:
Being applied in https://svn.apache.org/viewvc?view=revision&revision=1799399.

PR #12, opened 2017-07-28
JCRVLT-196 Add remapping support for other than renames

tripodsan:


tteofili:
I think package-info needs to be bumped org.apache.jackrabbit.vault.fs.io to 2.6.0 and org.apache.jackrabbit.vault.fs.api to 2.5.0

PR #15, opened 2017-12-02
Add feature to handover package properties to JCR Events.

joerghoh:
It's sometime useful if you can provide a flag/string in a content package, which can be evaluated in the JCR observation handlers using ```event.getUserData()```.

The flag must be set in the ```properties.xml``` like this

```
<entry key="eventUserData">myValue</entry>
```


tripodsan:
@joerghoh we decided not to do this, right? can you close this pr. thanks.

PR #18, opened 2017-12-04
JCRVLT-248 Support install hook bundles

tripodsan:


tripodsan:
closing due to inactivity.

PR #19, opened 2017-12-07
JCRVLT-251 test case for "Child nodes mentioned in parent node's docview

kwin:
xml are not cleared correctly during installation"


tripodsan:
@kwin the respective issue was closed. can you close this PR? thanks.

kwin:
@tripodsan If you want to have this behaviour for backwards-compatibility reasons, then I think it is reasonable to add the test but revert the expectations. WDYT?

PR #20, opened 2018-01-08
bugfix/JCRVLT-257: Check if environment supports compression level changes

Buuhuu:
This PR implements the check suggested here https://github.com/madler/zlib/issues/305 to verify if the change of the compression level for individual jar entries is supported by the running environment. 

For me (jdk 1.8.0 151 and jdk 9.0.1+11 with zlib 1.2.11) that fixes the issue.


Buuhuu:
Applied in r1820625

PR #21, opened 2018-02-01
Add rudimentary support for controlling the line separator

beaudet:
Requiring SVN settings to change in order to adapt to vlt's inability to control the line separator being used, combined with the surprising (fact?) that it's impossible to create a working windows batch file that passes chr(10) on the command line to -Dline.separator and the fact that it's a bit tricky to do with bash, led me to creating this patch that allows a custom vlt.line.separator system property to be set to a text value of either LF or CRLF which is then read in main() and set accordingly.  I realize that this approach would probably need to be adapted to conform with vlt's settings (or at least to accept a new command line option) but I didn't have the time to investigate that route.  I'm hoping the existence of this pull request and the utility it provides in giving teams control over the line separator so that the JCR, version control systems, and vlt can work harmoniously together regardless of which operating system is used.  For example, since vlt on Mac presumably defaults to LF, a team comprised of developers using Windows and Macs cannot all use vlt since they would be creating files that are detected by their version control system as being different due to the line separator differences that vlt creates across those systems.

### DIFF
--- a/vault-cli/src/main/appassembler/unix-template.sh
+++ b/vault-cli/src/main/appassembler/unix-template.sh
@@ -13,12 +13,13 @@
 #    e.g. to debug vault itself, use
 #      set VLT_OPTS=-Xdebug -Xnoagent -Djava.compiler=NONE -Xrunjdwp:transport=dt_socket,server=y,suspend=y,address=8000
 #
-#    LINE SEPARATOR
-#    To force vault to use a particular line separator regardless of the operating system's native line separator (default), you
-#    can inject -Dline.separator=$'\r'$'\n' or -Dline.separator=$'\n' into the exec line where VLT_OPTS is used; it might
-#    also be possible to embed that in the VLT_OPTS variable but that has proven problematic.  Perhaps VLT will support such a
-#    configuration as a native option in the future and call System.setProperties("line.separator",lsep) which also works
-#----------------------------------------------------------------------------
+# LINE SEPARATOR
+# Because it can be pretty tricky to shove a newline character into a batch or shell script and the text "\n" doesn't get
+# interpreted as what you might think by Java, vlt will instead look for LF | CRLF in a system property called vlt.line.separator
+# and if it exists, it will set the appropriate Java System property accordingly, e.g. -Dvlt.line.separator=LF
+# ----------------------------------------------------------------------------
+
+VLT_OPTS=-Dvlt.line.separator=LF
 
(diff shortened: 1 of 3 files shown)


beaudet:
The objective we're trying to achieve is to be able to use LF as the line terminator in all of our text files.  However, it seems that a checkout on a Windows system converts \n to \r\n.  When opening such a file in a windows tool that defaults to understanding LF as \n, e.g. vim under cygwin, we see the extraneous \r's. I don't see how setting the SVN property svn:eol-style=native will fix this.

For example, assuming SVN is set to native, this will force SVN to return CRLF line terminated files to Windows clients at checkout.  This is not what we want.  We want Windows to get LF line terminated files so that text editors such as vim running on cygwin can be used for editing for developers wishing to use that tool.

If we set eol-style to LF, then SVN will always return LF, but under this scenario, when we use VLT to check-out from the repository then perform an SVN diff on the checked-out files, we will see a bunch of changes pending due to VLT assuming CRLF line termination so we're back to the original problem. 

This is why the patch was developed.  So, unless I'm missing something like VLT reading the SVN configuration, I'm still confused about how using SVN settings would solve our problem.

tripodsan:
ok, now I understand - using cygwin and vim is problematic in this respect as it mixes the linefeed problematic. but do you really develop using vim under windows? also, since the svn:eol-style is defined on the resource itself, you would get in trouble with your coworkers that don't use cygwin. a better alternative would be to use git, where the linefeed behaviour is set by the user.

In any case, I can see your problem. please file an issue in the issue tracker:
https://issues.apache.org/jira/projects/JCRVLT

I think adding this to the batch/shell script is ok, but you should add support for environment variable, otherwise you have to patch your start script every time you update vault.
if we want to add this support to the cli, we would need to add it to `org.apache.jackrabbit.vault.cli.VaultFsApp#addApplicationOptions` as a global option. but I don't think this would be too useful as you would need to specify it for every command.

beaudet:
Yes, some of us use vim under cygwin to edit certain files.  It's a lot more convenient when you're using command line utilities to find the files of interest.  Regardless, I'll rework this pull request to support the environment variable approach.  It's important that a plain text value is used as the value for the variable since storing / passing the literal line separator is problematic, particularly in Windows bat files.  An issue was created here: https://issues.apache.org/jira/browse/JCRVLT-284

beaudet:
Thanks for committing the patch for this!

PR #23, opened 2018-02-06
Changing log level from error to info in case of jar file doesn't con…

nit23uec:
…tains jcr_root

@trivial
cc @tripodsan

### DIFF
--- a/vault-core/src/main/java/org/apache/jackrabbit/vault/packaging/registry/impl/JcrPackageRegistry.java
+++ b/vault-core/src/main/java/org/apache/jackrabbit/vault/packaging/registry/impl/JcrPackageRegistry.java
@@ -414,7 +414,7 @@ public class JcrPackageRegistry implements PackageRegistry {
 
         if (archive.getJcrRoot() == null) {
             String msg = "Stream is not a content package. Missing 'jcr_root'.";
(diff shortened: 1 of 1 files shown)


tripodsan:
can you give a reason why this change is good? as a user of the API, I would expect the appropriate error message shown if I upload a wrong package.

please open a jira ticket, if you still think this change is required.

nit23uec:
@tripodsan the jar file could contain some native components as well (c/c++ components compiled for specific platforms and in this case, they would not be OSGI bundles). They are meant to be processes using c custom resource transformer rather than a PackageTransformer which call jackrabbit layer. Also, the PackageTransformer gives info in the logs rather than the error which seems to be inconsistent with the jackrabbit layer logging.

nit23uec:
created https://issues.apache.org/jira/browse/JCRVLT-267

tripodsan:
this comment is not related to this apache project. please keep the communication channels clear.

From the point of this API, the package manager doesn't know about PackageTransformer, or potentially uploaded OSGi bundles. All it knows is, that it receives a stream that is not a package and therefore logs an error.

nit23uec:
Thanks for the guidance @tripodsan . I'm closing this PR.

PR #24, opened 2018-02-15
JCRVLT-271: formatting cli command

Buuhuu:


vault-core/src/main/java/org/apache/jackrabbit/vault/fs/io/DocViewFormat.java
```diff
@@ -0,0 +1,215 @@
...
+     */
+    public static void format(File file, List<Pattern> filenamePatterns) throws IOException {
+        Files.walkFileTree(file.toPath(), new AbstractFormattingVisitor(filenamePatterns) {
+            @Override protected void process(File file) throws IOException {
+                format(file);
+            }
+        });
+    }
+
+    private static synchronized byte[] format(File file, Checksum original, Checksum formatted) throws IOException {
```
tripodsan: why synchronized ?

vault-core/src/main/java/org/apache/jackrabbit/vault/fs/io/DocViewFormat.java
```diff
@@ -0,0 +1,215 @@
...
+import org.apache.jackrabbit.vault.util.xml.serialize.XMLSerializer;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+import org.xml.sax.InputSource;
+import org.xml.sax.SAXException;
+import org.xml.sax.XMLReader;
+import org.xml.sax.helpers.XMLReaderFactory;
+
+/**
+ * Singleton class providing access to the commonly used doc view xml format and functionality that checks files for the format or reformats
```
tripodsan: why singleton? I guess this is why it can't be used concurrently.

vault-core/src/test/java/org/apache/jackrabbit/vault/fs/io/DocViewFormatTest.java
```diff
@@ -0,0 +1,78 @@
...
+import java.util.Date;
+import java.util.List;
+import java.util.regex.Pattern;
+
+import org.apache.commons.io.IOUtils;
+import org.junit.After;
+import org.junit.Before;
+import org.junit.Test;
+
+public class DocViewFormatTest {
```
tripodsan: what about adding tests that check if the formatted output is correct?

vault-core/src/main/java/org/apache/jackrabbit/vault/util/xml/serialize/XMLSerializer.java
```diff
@@ -1100,6 +1109,73 @@ private void printNamespaceAttr(String prefix, String uri) throws IOException {
         _printer.printText('"');
     }
 
+    private List<Node> sortAttributes(NamedNodeMap attributeList) {
```
tripodsan: I dont like this change. if there is no comparator defined, you add the overhead of copying all attributes into a new list.

vault-core/src/main/java/org/apache/jackrabbit/vault/fs/io/DocViewFormat.java
Buuhuu: Mainly because of the buffer reused for formatted files. I made the methods non-static which makes sense for concurrency but also due to the fact that OutputFormat is mutable and though should not be hold statically.

vault-core/src/main/java/org/apache/jackrabbit/vault/fs/io/DocViewFormat.java
Buuhuu: Changed.

vault-core/src/main/java/org/apache/jackrabbit/vault/util/xml/serialize/XMLSerializer.java
Buuhuu: According to NamedNodeMap's javadoc 

> Objects contained in an object implementing NamedNodeMap may also be accessed by an ordinal index, but this is simply to allow convenient enumeration of the contents of a NamedNodeMap, and does not imply that the DOM specifies an order to these Nodes

the attributes in the NamedNodeMap don't imply an order. Accessing using the index is just for convenience. 

As the usage of the attributes is sequential access only, I changed that to use an Iterable around the NamedNodeMap when no Comparator is given, and the List as before otherwise.

vault-core/src/test/java/org/apache/jackrabbit/vault/fs/io/DocViewFormatTest.java
Buuhuu: Not sure what you mean by checking if the formatted output is correct. Do you mean correct in terms of line breaks, indention etc.? I haven't changed that so I assume that has coverage already. On the other hand what is tested here is:
- for a malformed file checkFormat() returns false
- format() returns without an exception and afterwards
- for the now well formed file checkFormat() returns true

Would it make sense to have the well formed version of the file to compare against after formatting?

PR #30, opened 2018-07-03
JCRVLT-180 implement filesystem based persistence layer

tripodsan:
This is a work-in-progress pull request to facilitate discussion and reviews.

### DIFF
--- a/vault-core/src/main/java/org/apache/jackrabbit/vault/packaging/JcrPackageManager.java
+++ b/vault-core/src/main/java/org/apache/jackrabbit/vault/packaging/JcrPackageManager.java
@@ -34,6 +34,7 @@ import org.apache.jackrabbit.vault.fs.api.ProgressTrackerListener;
 import org.apache.jackrabbit.vault.fs.api.WorkspaceFilter;
 import org.apache.jackrabbit.vault.fs.io.Archive;
 import org.apache.jackrabbit.vault.fs.io.ImportOptions;
+import org.apache.jackrabbit.vault.packaging.registry.PackageRegistry;
 import org.osgi.annotation.versioning.ProviderType;
 
 /**
@@ -393,4 +394,7 @@ public interface JcrPackageManager extends PackageManager {
      */
     @Nonnull
     List<JcrPackage> listPackages(@Nullable String group, boolean built) throws RepositoryException;
+    
+    @Nonnull
+    PackageRegistry getRegistry();
 }
--- a/vault-core/src/main/java/org/apache/jackrabbit/vault/packaging/registry/impl/FSPackageRegistry.java
+++ b/vault-core/src/main/java/org/apache/jackrabbit/vault/packaging/registry/impl/FSPackageRegistry.java
@@ -17,11 +17,22 @@
 package org.apache.jackrabbit.vault.packaging.registry.impl;
 
 import java.io.File;
+import java.io.FileInputStream;
+import java.io.FileOutputStream;
 import java.io.IOException;
 import java.io.InputStream;
+import java.io.ObjectInputStream;
+import java.io.ObjectOutputStream;
+import java.io.Serializable;
+import java.util.ArrayList;
+import java.util.Collection;
 import java.util.Collections;
+import java.util.HashMap;
+import java.util.HashSet;
 import java.util.LinkedList;
 import java.util.List;
+import java.util.Map;
+import java.util.Map.Entry;
 import java.util.Properties;
 import java.util.Set;
 import java.util.TreeSet;
@@ -30,11 +41,13 @@ import java.util.UUID;
 import javax.annotation.Nonnull;
 import javax.annotation.Nullable;
 import javax.jcr.RepositoryException;
-import javax.jcr.Session;import org.apache.jackrabbit.vault.fs.config.MetaInf;
+import javax.jcr.Session;
+
+import org.apache.commons.io.FileUtils;
+import org.apache.jackrabbit.vault.fs.config.MetaInf;
 import org.apache.jackrabbit.vault.fs.io.ImportOptions;
 import org.apache.jackrabbit.vault.fs.io.MemoryArchive;
 import org.apache.jackrabbit.vault.packaging.Dependency;
-import org.apache.jackrabbit.vault.packaging.JcrPackage;
 import org.apache.jackrabbit.vault.packaging.NoSuchPackageException;
 import org.apache.jackrabbit.vault.packaging.PackageException;
 import org.apache.jackrabbit.vault.packaging.PackageExistsException;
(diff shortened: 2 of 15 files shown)


vault-core/src/main/java/org/apache/jackrabbit/vault/packaging/registry/impl/FSPackageRegistry.java
```diff
@@ -0,0 +1,570 @@
...
+                }
+            } finally {
+                if (pkg != null && !pkg.isClosed()) {
+                    pkg.close();
+                }
+            }
+
+        }
+         
+        // can't go through JcrPackageImpl as being JCR dependant but has to
```
tripodsan: remove unused code

vault-core/src/main/java/org/apache/jackrabbit/vault/packaging/registry/impl/FSPackageRegistry.java
```diff
@@ -0,0 +1,570 @@
...
+     * @param roots
+     *            the root paths to store the packages.
+     * @throws IOException
+     */
+    @SuppressWarnings("unchecked")
+    public FSPackageRegistry(@Nonnull File homeDir) throws IOException {
+        FSPackageRegistry.homeDir = homeDir;
+        if (homeDir.exists()) {
+            FSPackageRegistry.statusFile = new File(homeDir, "registrystatus.ser");
+            if (statusFile.exists()) {
```
tripodsan: move the status loading to own method

vault-core/src/main/java/org/apache/jackrabbit/vault/packaging/registry/impl/FSPackageRegistry.java
```diff
@@ -0,0 +1,570 @@
...
+                FileInputStream fis = null;
+                ObjectInputStream ois = null;
+                try {
+                    fis = new FileInputStream("hashmap.ser");
```
tripodsan: instead of 1 serialized hashmap, I think it's more robust to have a status file (properties or json) for each package. the file traversal overhead is not huge if it only happens once when a registry is created (assuming it can be cached as a service, for example). having individual files reduces concurrency issues.

vault-core/src/main/java/org/apache/jackrabbit/vault/packaging/registry/impl/FSPackageRegistry.java
```diff
@@ -0,0 +1,570 @@
...
+            return null;
+        }
+
+    }
+
+    /**
+     * {@inheritDoc}
+     */
+    @Override
+    public PackageId resolve(Dependency dependency, boolean onlyInstalled) throws IOException {
```
tripodsan: I think this could go into a common base class.

vault-core/src/main/java/org/apache/jackrabbit/vault/packaging/registry/impl/FSPackageRegistry.java
```diff
@@ -0,0 +1,570 @@
...
+        // TODO @suess - how to capture installation state
+        return false;
+    }
+
+    /**
+     * {@inheritDoc}
+     */
+    @Nonnull
+    @Override
+    public PackageId[] usage(PackageId id) throws IOException {
```
tripodsan: I think this could go into a common base class.

vault-core/src/main/java/org/apache/jackrabbit/vault/packaging/registry/impl/FSPackageRegistry.java
```diff
@@ -0,0 +1,570 @@
...
+                    }
+                }
+            }
+        }
+        return usages.toArray(new PackageId[usages.size()]);
+    }
+
+    @Nonnull
+    @Override
+    public DependencyReport analyzeDependencies(@Nonnull PackageId id, boolean onlyInstalled)
```
tripodsan: I think this could go into a common base class.

vault-core/src/main/java/org/apache/jackrabbit/vault/packaging/registry/impl/FSPackageRegistry.java
```diff
@@ -0,0 +1,570 @@
...
+            PackagePropertiesImpl props = new PackagePropertiesImpl() {
+                @Override
+                protected Properties getPropertiesMap() {
+                    return inf.getProperties();
+                }
+            };
+            PackageId pid = props.getId();
+    
+            // invalidate pid if path is unknown
+            if (pid == null) {
```
tripodsan: this is rather for backwards compatibility. I wonder if we should still allow this...

vault-core/src/main/java/org/apache/jackrabbit/vault/packaging/registry/impl/FSPackageRegistry.java
```diff
@@ -0,0 +1,570 @@
...
+        }
+        FileUtils.copyFile(file, pkgFile);
+        return pack.getId();
+    }
+
+    @Nonnull
+    @Override
+    public PackageId registerExternal(@Nonnull File file, boolean replace) throws IOException, PackageExistsException {
+        ZipVaultPackage pack = new ZipVaultPackage(file, false, true);
+        // TODO: @suess - capture reference in central location
```
tripodsan: or maybe add a `<package-path>.ref` to where the .zip file would be and just record the reference to the external file.

or just have a `<package-path>.properties`  as mandatory file that contains installation state etc. it could contain a `fileReference` property, in case the `<package-path>.zip` does not exist.

vault-core/src/main/java/org/apache/jackrabbit/vault/packaging/registry/impl/FSPackageRegistry.java
```diff
@@ -0,0 +1,570 @@
...
+            if (v.length() > 0) {
+                b.append("-").append(v);
+            }
+            installPath = b.toString();
+        }
+
+        return installPath;
+    }
+
+    private String getMappedPackage(PackageId id) {
```
tripodsan: what is this?

vault-core/src/main/java/org/apache/jackrabbit/vault/packaging/registry/impl/FSPackageRegistry.java
```diff
@@ -0,0 +1,570 @@
...
+    @Override
+    public ExecutionPlanBuilder createExecutionPlan() {
+        return new ExecutionPlanBuilderImpl(this);
+    }
+
+    @Override
+    public void installPackage(Session session, RegisteredPackage pkg, ImportOptions opts, boolean extract) throws IOException, PackageException {
+
+        if (session == null){
+            log.error("Installation of packages only supported when session is set.");
```
tripodsan: maybe throw an exception here?

vault-core/src/main/java/org/apache/jackrabbit/vault/packaging/registry/impl/FSPackageRegistry.java
DominikSuess: Not sure.. I think we could also throw an IOException and no longer support it.

vault-core/src/main/java/org/apache/jackrabbit/vault/packaging/JcrPackageManager.java
```diff
@@ -34,6 +34,7 @@
 import org.apache.jackrabbit.vault.fs.api.WorkspaceFilter;
 import org.apache.jackrabbit.vault.fs.io.Archive;
 import org.apache.jackrabbit.vault.fs.io.ImportOptions;
+import org.apache.jackrabbit.vault.packaging.registry.PackageRegistry;
```
tripodsan: remove unused import

vault-core/src/main/java/org/apache/jackrabbit/vault/packaging/registry/impl/FSPackageRegistry.java
```diff
@@ -0,0 +1,528 @@
...
+    public boolean contains(@Nonnull PackageId id) throws IOException {
+        File meta = getPackageMetaDataFile(id);
+        return meta.exists() && meta.isFile();
+    }
+
+    @Nullable
+    private File getPackageFile(@Nonnull PackageId id) {
+        try {
+            InstallState state = getInstallState(id);
+            if (FSPackageStatus.NOTREGISTERED.equals(state.getStatus())) {
```
tripodsan: enums can be compared with `==`

vault-core/src/main/java/org/apache/jackrabbit/vault/packaging/registry/impl/FSPackageRegistry.java
```diff
@@ -0,0 +1,528 @@
...
+        setInstallState(pid, targetStatus, state.getFilePath(), state.isExternal());
+    }
+
+    private void setInstallState(PackageId pid, FSPackageStatus targetStatus, String filePath, boolean external) throws IOException {
+        File metaData = getPackageMetaDataFile(pid);
+        
+        if (targetStatus.equals(FSPackageStatus.NOTREGISTERED)) {
+            metaData.delete();
+        } else {
+            OutputStream out = FileUtils.openOutputStream(metaData);
```
tripodsan: are you sure you want to use XML? isn't json or properties much easier? (and potentially faster) ?

vault-core/src/main/java/org/apache/jackrabbit/vault/packaging/registry/impl/FSPackageRegistry.java
```diff
@@ -0,0 +1,528 @@
...
+ * FileSystem based registry not depending on a JCR Session. All metadata is stored in Filesystem and can be prepared and used without a running JCR repository
+ * Only methods to install or uninstall packages require an active {@code Session} object of running jcr instance to perform the actual installation tasks 
+ */
+@Component(
+        service = PackageRegistry.class,
+        configurationPolicy = ConfigurationPolicy.REQUIRE,
+        property = {"service.vendor=The Apache Software Foundation"}
+)
+@Designate(ocd = FSPackageRegistry.Config.class)
+public class FSPackageRegistry extends AbstractPackageRegistry {
```
tripodsan: I think this should eventually to into some generic `getRegistry()` method in the `Packaging` class.

vault-core/src/main/java/org/apache/jackrabbit/vault/packaging/registry/impl/FSPackageRegistry.java
DominikSuess: @tripodsan  reason for xml is just that I wanted to reuse the same serialization  in vault-core already used for execution plan and didn#t want to introduce yet another serializer

vault-core/src/main/java/org/apache/jackrabbit/vault/packaging/registry/impl/FSPackageRegistry.java
DominikSuess: @tripodsan  the trouble is that we have active registries at the same time, but yes we certainly can improve this - but we should think on how to declare which registry to consume!?

vault-core/src/main/java/org/apache/jackrabbit/vault/packaging/registry/impl/FSPackageRegistry.java
```diff
@@ -0,0 +1,804 @@
...
+    )
+    @interface Config {
+        
+        @AttributeDefinition
+        String homePath() default "packageregistry";
+    }
+
+    @Activate
+    private void activate(BundleContext context, Config config) {
+        String repoHome = context.getProperty(REPOSITORY_HOME);
```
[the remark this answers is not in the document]
DominikSuess: done

PR #31, opened 2018-08-01
JCRVLT-313 - making FSRegisteredPackage build on cachable FSInstallSt…

DominikSuess:
…ate and return vltpackage only lazily


tripodsan:
@DominikSuess this is merged in. can you close it?

PR #32, opened 2018-08-01
Issue/jcrvlt 314

DominikSuess:


vault-core/src/main/java/org/apache/jackrabbit/vault/packaging/registry/impl/FSPackageRegistry.java
```diff
@@ -231,7 +231,7 @@ private File getPackageMetaDataFile(@Nonnull PackageId id) {
     /**
      * {@inheritDoc}
      */
-    public VaultPackage open(File pkg) throws RepositoryException {
+    public VaultPackage open(File pkg) throws IOException {
```
tripodsan: I would not return NULL here, as the consumer (FSRegisteredPackage) doesn't expect this.

vault-core/src/main/java/org/apache/jackrabbit/vault/packaging/registry/impl/FSRegisteredPackage.java
```diff
@@ -39,24 +39,27 @@
...
-        return vltPkg.getId();
+        return installState.getPackageId();
     }
 
     @Nonnull
     @Override
     public VaultPackage getPackage() throws IOException {
-        return vltPkg;
+        this.vltPkg = registry.open(installState.getFilePath().toFile());
+        return this.vltPkg;
```
tripodsan: you overwrite `this.vltPkg`. is this intended? or would it be better to do:

```
if (this.vltPkg == null) {
  this.vltPkg = ...
}
```

vault-core/src/main/java/org/apache/jackrabbit/vault/packaging/registry/impl/FSRegisteredPackage.java
```diff
@@ -39,24 +39,27 @@
...
-    private VaultPackage vltPkg;
+    private FSInstallState installState;
     private FSPackageRegistry registry;
 
-    public FSRegisteredPackage(FSPackageRegistry registry, VaultPackage vltPkg) throws IOException, RepositoryException {
-        this.vltPkg = vltPkg;
+    private VaultPackage vltPkg = null;
+
+    public FSRegisteredPackage(FSPackageRegistry registry, FSInstallState installState) throws IOException, RepositoryException {
+        this.installState = installState;
```
tripodsan: what about the other methods like `isInstalled()` or `getInstallationTime()`. shouldn't they rely on the immutable installstate?

otherwise, I would not store the installState in this object, but only remember `packageId` and `filePath`. i..e:
```
this.packageId = installState.getPackageId();
this.filePath = installState.getFilePath();
```

tripodsan:
@DominikSuess this PR is merged into trunk. can you close it? thanks.

PR #36, opened 2018-09-06
JCRVLT-319 - allow definition of packageIds that are externally satis…

DominikSuess:
…fied when building executinplan - consumer takes over responsibility to install those upfront (allows building of stacked executionplans)

### DIFF
--- a/vault-core/src/main/java/org/apache/jackrabbit/vault/packaging/registry/ExecutionPlanBuilder.java
+++ b/vault-core/src/main/java/org/apache/jackrabbit/vault/packaging/registry/ExecutionPlanBuilder.java
@@ -103,7 +103,7 @@ public interface ExecutionPlanBuilder {
      * @throws PackageException if the plan is not valid.
      */
     @Nonnull
-    List<PackageId> calculate() throws IOException, PackageException;
+    List<PackageId> calculateIds() throws IOException, PackageException;
     
     /**
      * builds an executes the plan synchronously.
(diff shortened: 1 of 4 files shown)


vault-core/src/main/java/org/apache/jackrabbit/vault/packaging/registry/ExecutionPlanBuilder.java
```diff
@@ -84,6 +86,25 @@
...
+
+    /**
+     * Triggers Validation and returns PackageIds of all packages to be installed
+     * by this builder
+     * @return List of packages to be installed by this builder.
+     * @throws IOException if an I/O error occurrs.
+     * @throws PackageException if the plan is not valid.
+     */
+    @Nonnull
+    List<PackageId> calculateIds() throws IOException, PackageException;
```
tripodsan: - Should be a Set (unless the order is important)
- I don't like the name. The tasks could also be removal tasks. so what about `List<PackageTask> preview()` ?

DominikSuess: fixed

DominikSuess:
@tripodsan  this somehow fell through the cracks - as I'm just stuck due to this as I currently don't get clean layering of execution plans without this. it would be great if you could have a quick look - I 
 incorporated your proposed changes.

PR #38, opened 2018-10-23
JCRVLT-141 Log all activity to debug so that an activity log can be c…

vaibhav2601:
…reated


tripodsan:
please combine with #37

PR #41, opened 2019-03-06
JCRVLT-332 - adding support to JcrPackageRegistry to resolve dependen…

DominikSuess:
…cies from FsPackageRegistry

### DIFF
--- a/vault-core/src/main/java/org/apache/jackrabbit/vault/packaging/registry/impl/JcrPackageRegistry.java
+++ b/vault-core/src/main/java/org/apache/jackrabbit/vault/packaging/registry/impl/JcrPackageRegistry.java
@@ -250,9 +250,7 @@ public class JcrPackageRegistry extends AbstractPackageRegistry {
         try {
             Node node = getPackageNode(id);
             if (node == null && baseRegistry != null) {
-                if (baseRegistry.contains(id)) {
-                    return baseRegistry.open(id);
-                }
+                return baseRegistry.open(id);
             }
             return node == null ? null : new JcrRegisteredPackage(open(node, false));
         } catch (RepositoryException e) {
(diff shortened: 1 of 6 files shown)


vault-core/src/main/java/org/apache/jackrabbit/vault/packaging/registry/impl/JcrPackageRegistry.java
```diff
@@ -115,6 +115,11 @@
      * the package root prefix of the primary root path.
      */
     private final String primaryPackRootPathPrefix;
+    
+    /**
+     * FSPackageRegistry can be registered if present in the system to be able to look up presatisfied dependencies
+     */
+    private FSPackageRegistry fsPackageRegistry = null;
```
tripodsan: does this need to be a `FSPackageRegistry` or could this just be a _base_ or _fallback_ package registry?

```suggestion
    private PackageRegistry baseRegistry = null;
```

vault-core/src/main/java/org/apache/jackrabbit/vault/packaging/registry/impl/JcrPackageRegistry.java
```diff
@@ -241,6 +249,11 @@ public Node getPrimaryPackageRoot(boolean autoCreate) throws RepositoryException
     public RegisteredPackage open(@Nonnull PackageId id) throws IOException {
         try {
             Node node = getPackageNode(id);
+            if (node == null && baseRegistry != null) {
+                if (baseRegistry.contains(id)) {
```
tripodsan: can't this be simplified to just:

```suggestion
                return baseRegistry.open(id)
```

or does it bring additional value to check `contains()` first?

DominikSuess:
public JcrPackage open(Node node, boolean allowInvalid);
won't work as Registry doesn't return a JcrPackage but a RegisteredPackage. Contains I'll be adjusting

DominikSuess:
Pushed requested changes but the open method of JcrPackageRegistry which explicitly opens a JcrPackage which the Registry Interface is not aware of (and in FSRegistry wouldn't exist)

PR #42, opened 2019-03-14
JCRVLT-332 - fixing osgi reference (dynamic reference policy & volati…

DominikSuess:
…le field) to work under production conditions


tripodsan:
thanks...

PR #43, opened 2019-04-01
JCRVLT-336 - adding support for configuring installationScope for FSP…

DominikSuess:
…ackageRegistry

### DIFF
--- a/vault-core/src/main/java/org/apache/jackrabbit/vault/packaging/registry/impl/FSPackageRegistry.java
+++ b/vault-core/src/main/java/org/apache/jackrabbit/vault/packaging/registry/impl/FSPackageRegistry.java
@@ -116,7 +116,7 @@ public class FSPackageRegistry extends AbstractPackageRegistry {
 
     private File homeDir;
 
-    private String scope;
+    private InstallationScope scope;
 
     private File getHomeDir() {
         return homeDir;
@@ -129,7 +129,7 @@ public class FSPackageRegistry extends AbstractPackageRegistry {
      * @throws IOException If an I/O error occurs.
      */
     public FSPackageRegistry(@Nonnull File homeDir) throws IOException {
-        this(homeDir, InstallationScope.UNSCOPED);
+        this(homeDir, new InstallationScope(InstallationScope.UNSCOPED));
     }
 
     /**
@@ -139,7 +139,7 @@ public class FSPackageRegistry extends AbstractPackageRegistry {
      * @param scope to set a corresponding workspacefilter
      * @throws IOException If an I/O error occurs.
      */
-    public FSPackageRegistry(@Nonnull File homeDir, String scope) throws IOException {
+    public FSPackageRegistry(@Nonnull File homeDir, InstallationScope scope) throws IOException {
         this.homeDir = homeDir;
         this.scope = scope;
         loadPackageCache();
@@ -187,7 +187,7 @@ public class FSPackageRegistry extends AbstractPackageRegistry {
                 homeDir.mkdirs();
             }
         }
-        this.scope = config.scope();
+        this.scope = new InstallationScope(config.scope());
         loadPackageCache();
         log.info("Jackrabbit Filevault FS Package Registry initialized with home location {}", this.homeDir.getPath());
     }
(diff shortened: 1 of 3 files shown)


vault-core/src/main/java/org/apache/jackrabbit/vault/packaging/registry/impl/FSPackageRegistry.java
```diff
@@ -123,7 +129,19 @@ private File getHomeDir() {
...
+    }
+
+    /**
+     * Creates a new FSPackageRegistry based on the given home directory.
+     *
+     * @param homeDir the directory in which packages and their metadata is stored
+     * @param scope to set a corresponding workspacefilter
+     * @throws IOException If an I/O error occurs.
+     */
+    public FSPackageRegistry(@Nonnull File homeDir, String scope) throws IOException {
```
tripodsan: why not use the `InstallationScope` class ?

vault-core/src/main/java/org/apache/jackrabbit/vault/packaging/registry/impl/InstallationScope.java
```diff
@@ -0,0 +1,23 @@
...
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.apache.jackrabbit.vault.packaging.registry.impl;
+
+public final class InstallationScope {
```
tripodsan: wouldn't a `enum` be better ?

vault-core/src/main/java/org/apache/jackrabbit/vault/packaging/registry/impl/FSPackageRegistry.java
```diff
@@ -112,6 +116,8 @@
 
     private File homeDir;
 
+    private String scope;
```
tripodsan: use `InstallationScope` class

DominikSuess: see comment above.. about to resolve.

vault-core/src/main/java/org/apache/jackrabbit/vault/packaging/registry/impl/InstallationScope.java
DominikSuess: @tripodsan  - actually I started with it but the problem is that there is no static access to the string values of an enum making this unusable for the configuration case - I'll switch that to a normal POJO with constants and getters/setters for a single value so I can use it in the internals to pass not string objects.  Patched now

vault-core/src/main/java/org/apache/jackrabbit/vault/packaging/registry/impl/InstallationScope.java
```diff
@@ -0,0 +1,41 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.apache.jackrabbit.vault.packaging.registry.impl;
+
+import java.util.Arrays;
+
+public class InstallationScope {
```
tripodsan: why not enum?

vault-core/src/main/java/org/apache/jackrabbit/vault/packaging/registry/impl/InstallationScope.java
tripodsan: `InstallationScope.UNSCOPED.name()` ?

DominikSuess: .name() is also not a constant expression (as it may be overridden) - getting corresponding errors.

vault-core/src/main/java/org/apache/jackrabbit/vault/packaging/registry/impl/InstallationScope.java
DominikSuess: @tripodsan  - see above, there is no way to access the values of the enum as a String with a static expression which I need for the annotations at build time.

tripodsan: well, I think that having the benefit of the `enum` throughout the code outweighs to use 3 strings in the config :-)

PR #45, opened 2019-04-04
JCRVLT-337 - pick up WorkspaceFilter override from ImportOptions and …

DominikSuess:
…respect filter in ExportOptions when creating snapshot


DominikSuess:
@tripodsan  - patch proposal - corresponding test case pending for a scoped filter example

kwin:
Something is fishy with this PR, 688 changed files seems wrong. @DominikSuess Can  you rebase?

PR #48, opened 2019-05-29
JCRVLT-338 : argLine getting overridden by parent pom and vault-core pom

sonagupt:
argLine needs to be appended rather than overridden.
Not able to install jacoco via cli as the argLine is getting overridden.

### DIFF
--- a/parent/pom.xml
+++ b/parent/pom.xml
@@ -45,6 +45,7 @@
         <slf4j.version>1.7.6</slf4j.version>
         <test.oak>false</test.oak>
         <test.vm.options>-Doak=$test.oak}</test.vm.options>
+        <argLine>${test.vm.options}</argLine>
     </properties>
 
     <!-- ====================================================================== -->
@@ -172,7 +173,6 @@
                                 <value>target/derby.log</value>
                             </property>
                         </systemProperties>
-                        <argLine>${argLine} ${test.vm.options}</argLine>
                     </configuration>
                 </plugin>
                 <!-- ====================================================================== -->
--- a/parent/pom.xml
+++ b/parent/pom.xml
@@ -172,7 +172,7 @@
                                 <value>target/derby.log</value>
                             </property>
                         </systemProperties>
-                        <argLine>${test.vm.options}</argLine>
+                        <argLine>${argLine} ${test.vm.options}</argLine>
                     </configuration>
                 </plugin>
                 <!-- ====================================================================== -->


sonagupt:
@tripodsan Please review

parent/pom.xml
```diff
@@ -45,6 +45,7 @@
         <slf4j.version>1.7.6</slf4j.version>
         <test.oak>false</test.oak>
         <test.vm.options>-Doak=$test.oak}</test.vm.options>
+        <argLine>${test.vm.options}</argLine>
```
tripodsan: I'm not sure this is correct. the argline might be used by other plugins as well. I'd rather keep it with the surefire plugin.

tripodsan:
maybe you can provide an example on how you want to invoke the filevault build from your jacoco cli?

sonagupt:
I want to use command : mvn clean org.jacoco:jacoco-maven-plugin:0.8.3:prepare-agent verify org.jacoco:jacoco-maven-plugin:0.8.3:report
Please refer : https://github.com/Alfresco/alfresco-sdk/issues/298
https://stackoverflow.com/questions/46489455/append-the-value-of-argline-param-in-maven-surefire-plugin
https://github.com/jacoco/jacoco/issues/597

tripodsan:
@sonagupt and what is the `argLine` you want to pass into the surefire pluigin? i.e. if you would do it manually, what would you change?

sonagupt:
@tripodsan The argLine for a maven project is common for all the plug-ins. All plugins send arguments which are used while executing maven install. So here surefire plugin is overriding the argLine hence I am not able to append anything using jacoco plugin. As mentioned in the articles the best practice is that argLine parameter should be defined as a property of the pom, rather than a configuration of surefire.
Please refer : Alfresco/alfresco-sdk#298
https://stackoverflow.com/questions/46489455/append-the-value-of-argline-param-in-maven-surefire-plugin
jacoco/jacoco#597

tripodsan:
@sonagupt I understand. but please, share you configuration or how you intend to invoke the filevault build.

sonagupt:
@tripodsan :  mvn clean org.jacoco:jacoco-maven-plugin:0.8.3:prepare-agent verify org.jacoco:jacoco-maven-plugin:0.8.3:report

tripodsan:
fixed in https://github.com/apache/jackrabbit-filevault/commit/2ad8f61143a88f3bbc4fe693cccf1888adc2f4c5

@sonagupt please close this PR for me, thanks.

PR #52, opened 2019-08-16
JCRVLT-120 document property filters

kwin:
extend information about import/installation


kwin:
Merged via http://svn.apache.org/r1865905.

PR #53, opened 2019-08-26
JCRVLT-356 add equals and toString to DocViewNode and DocViewProperty

kwin:
migrate unit tests to JUnit4


kwin:
Merged in http://svn.apache.org/r1865971.

PR #54, opened 2019-08-28
SKYOPS-2057 rep:cugPolicy node not replicated on Skyline instances

akrainiouk:
- cugPolicy nodes are treated as regular nodes so they are not affected by aclHandling configuration option.


vault-core/src/main/java/org/apache/jackrabbit/vault/fs/impl/io/DocViewSAXImporter.java
```diff
@@ -627,7 +630,7 @@ public void startElement(final String uri, final String localName, final String
                 } else {
                     try {
                         DocViewNode ni = new DocViewNode(name, label, attributes, npResolver);
-                        if (aclManagement.isACLNodeType(ni.primary)) {
+                        if (isACLNode(ni.primary)) {
```
tripodsan: don't hardcode this here...move it to `aclManagement`

PR #56, opened 2019-09-01
JCRVLT-359 Provide separate option to control handling of rep:cugPoli…

akrainiouk:
…cy nodes.

- cugHandling option added to ImportOptions. If non null, then cugPolicy handling is controlled by this option rather then aclHandling.


akrainiouk:
> LGTM, but please add some tests.

Yes, I am working on it but it is a long weekend in US and I wanted you guys to take a look in the mean time.

PR #57, opened 2019-09-03
JCRVLT-362 correctly expose input stream also for META-INF entries

kwin:


kwin:
Fixed by https://svn.apache.org/viewvc?view=revision&revision=1866356

PR #59, opened 2019-10-04
Feature/jcrvlt 382 correctly reset source

kwin:


kwin:
Fixed with http://svn.apache.org/r1868224.

PR #60, opened 2019-10-10
Bugfix/jcrvlt 368 throw proper exception for invalid pattern

kwin:


kwin:
Fixed with http://svn.apache.org/r1868470.

PR #63, opened 2019-10-20
Feature/jcrvlt 371 dedicated property for dep location

kwin:


kwin:
Merged in http://svn.apache.org/r1868686.

PR #64, opened 2019-11-11
JCRVLT-419: Feature/enable insecure https host

wildone:
Provide insecure flag to enable access to https with expired certs.

Adding a new cli argument `--insecure`.

Enabling optional support for expired ssl certs when using https on development server with self generated certificates.

There is a corresponding PR in the required jackrabbit dependency apache/jackrabbit#88

### DIFF
--- a/vault-cli/src/main/java/org/apache/jackrabbit/vault/util/console/AbstractApplication.java
+++ b/vault-cli/src/main/java/org/apache/jackrabbit/vault/util/console/AbstractApplication.java
@@ -210,17 +210,6 @@ public abstract class AbstractApplication {
                         )
                         .create();
 
-        optInsecureHttps =
-                obuilder
-                        .withLongName("insecure")
-                        .withDescription("allow expired ssl certs for https")
-                        .withArgument(abuilder
-                                .withName("command")
-                                .withMaximum(1)
-                                .create()
-                        )
-                        .create();
-
         gbuilder
                 .withName("Global options:")
                 //.withOption(optPropertyFile)
--- a/vault-cli/src/main/java/org/apache/jackrabbit/vault/util/console/AbstractApplication.java
+++ b/vault-cli/src/main/java/org/apache/jackrabbit/vault/util/console/AbstractApplication.java
@@ -210,6 +210,17 @@ public abstract class AbstractApplication {
                         )
                         .create();
 
+        optInsecureHttps =
+                obuilder
+                        .withLongName("insecure")
+                        .withDescription("allow expired ssl certs for https")
+                        .withArgument(abuilder
+                                .withName("command")
+                                .withMaximum(1)
+                                .create()
+                        )
+                        .create();
+
         gbuilder
                 .withName("Global options:")
                 //.withOption(optPropertyFile)
--- a/vault-cli/src/main/java/org/apache/jackrabbit/vault/util/console/AbstractApplication.java
+++ b/vault-cli/src/main/java/org/apache/jackrabbit/vault/util/console/AbstractApplication.java
@@ -16,13 +16,6 @@
  */
 package org.apache.jackrabbit.vault.util.console;
 
-import java.io.File;
-import java.io.FileInputStream;
-import java.io.FileOutputStream;
-import java.io.IOException;
-import java.util.Iterator;
-import java.util.Properties;
-
 import org.apache.commons.cli2.CommandLine;
 import org.apache.commons.cli2.DisplaySetting;
 import org.apache.commons.cli2.Group;
@@ -39,6 +32,13 @@ import org.apache.jackrabbit.vault.util.console.util.Log4JConfig;
 import org.apache.jackrabbit.vault.util.console.util.PomProperties;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
+
+import java.io.File;
+import java.io.FileInputStream;
+import java.io.FileOutputStream;
+import java.io.IOException;
+import java.util.Iterator;
+import java.util.Properties;
 
 /**
  * {@code Console}...
@@ -69,6 +69,7 @@ public abstract class AbstractApplication {
     private Option optLogLevel;
     private Option optVersion;
     private Option optHelp;
+    private Option optInsecureHttps;
 
     public String getVersion() {
         return getPomProperties().getVersion();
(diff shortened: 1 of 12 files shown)


vault-cli/src/main/java/org/apache/jackrabbit/vault/util/console/AbstractApplication.java
```diff
@@ -198,6 +199,28 @@ public GroupBuilder addApplicationOptions(GroupBuilder gbuilder) {
...
+                        .withLongName("insecure")
+                        .withDescription("allow expired ssl certs for https")
+                        .withArgument(abuilder
+                                .withName("command")
+                                .withMaximum(1)
+                                .create()
+                        )
+                        .create();
+
+        optInsecureHttps =
```
tripodsan: why twice?

vault-core/src/main/java/org/apache/jackrabbit/vault/packaging/PackageId.java
```diff
@@ -209,14 +209,16 @@ public PackageId(String group, String name, String version) {
     public PackageId(String group, String name, Version version) {
         fromPath = false;
         // validate group
-        if (group.equals(ETC_PACKAGES)) {
+        if (group == null || group.equals(ETC_PACKAGES)) {
```
tripodsan: this change is not related to the issue.

vault-core/src/main/java/org/apache/jackrabbit/vault/packaging/PackageId.java
```diff
@@ -209,14 +209,16 @@ public PackageId(String group, String name, String version) {
...
             group = "";
         } else if (group.startsWith(ETC_PACKAGES_PREFIX)) {
             group = group.substring(ETC_PACKAGES_PREFIX.length());
         } else if (group.startsWith("/")) {
             group = group.substring(1);
         }
         this.group = group;
+        if (name == null)
```
tripodsan: this change is not related to the issue (and uses wrong coding convention)

vault-cli/src/main/java/org/apache/jackrabbit/vault/util/console/AbstractApplication.java
wildone: pebcak, removing

vault-core/src/main/java/org/apache/jackrabbit/vault/packaging/PackageId.java
wildone: removing

vault-core/src/main/java/org/apache/jackrabbit/vault/packaging/PackageId.java
wildone: removing, will make separate ticket for this

wildone:
> * why do you add more extra dependencies?

In a clean environment I was not able to compile this without these dependencies

> * please don't reformat the code

should put the whitespace back or is it ok? happy to not do it in future

tripodsan:
> should put the whitespace back or is it ok? happy to not do it in future

white space is ok, but rearranging imports is kind of confusing.

tripodsan:
> > * why do you add more extra dependencies?
> 
> In a clean environment I was not able to compile this without these dependencies

it looks like your environment doesn't load the transitive dependencies.... which is weird.
but can you anyways remove them from the PR? thanks.

wildone:
> > > * why do you add more extra dependencies?
> > 
> > 
> > In a clean environment I was not able to compile this without these dependencies
> 
> it looks like your environment doesn't load the transitive dependencies.... which is weird.
> but can you anyways remove them from the PR? thanks.

Ill update will see if it works

wildone:
@tripodsan ive removed those but looks like Travis is getting same issue that I am getting, can't compile anymore at all. The initial issue for me was that I did not have any of those dependencies in my .m2 repo so it would not build, I use github actions without .m2 cache so will see if it will build once existing version issues are updated.

tripodsan:
@wildone no, there is another build error:

```
[INFO]   PACKAGE_NAME                                       DELTA      CUR_VER    BASE_VER   REC_VER    WARNINGS  
[INFO] = ================================================== ========== ========== ========== ========== ==========
[INFO]   org.apache.jackrabbit.vault.fs                     unchanged  2.4.0      2.4.0      2.4.0      -         
[INFO] -----------------------------------------------------------------------------------------------------------
[INFO] * org.apache.jackrabbit.vault.fs.api                 major      2.7.0      2.7.0      3.0.0      Version increase required
[INFO]      > interface org.apache.jackrabbit.vault.fs.api.RepositoryFactory
[INFO]          + method createRepository(org.apache.jackrabbit.vault.fs.api.RepositoryAddress,boolean)
[INFO]              + access abstract
[INFO]              + return javax.jcr.Repository
```

since you changed the api, you would need to increase the major version of the api.

wildone:
> @wildone no, there is another build error:
> 
> ```
> [INFO]   PACKAGE_NAME                                       DELTA      CUR_VER    BASE_VER   REC_VER    WARNINGS  
> [INFO] = ================================================== ========== ========== ========== ========== ==========
> [INFO]   org.apache.jackrabbit.vault.fs                     unchanged  2.4.0      2.4.0      2.4.0      -         
> [INFO] -----------------------------------------------------------------------------------------------------------
> [INFO] * org.apache.jackrabbit.vault.fs.api                 major      2.7.0      2.7.0      3.0.0      Version increase required
> [INFO]      > interface org.apache.jackrabbit.vault.fs.api.RepositoryFactory
> [INFO]          + method createRepository(org.apache.jackrabbit.vault.fs.api.RepositoryAddress,boolean)
> [INFO]              + access abstract
> [INFO]              + return javax.jcr.Repository
> ```
> 
> you added a new method here:
> 
> https://github.com/apache/jackrabbit-filevault/pull/64/files#diff-ea5aa6d8c0d2ae0a75bab5a99988de7bR35
> 
> and now it requests you to increase the version....
> 
> The problem is, that the `RepositoryFactory` should be annotated with `@Provider`, so it only creates a minor version increase...
> 
> I can take a look....

that would be great thank you! :D

wildone:
is there anything I can help with @tripodsan ?

tripodsan:
> is there anything I can help with @tripodsan ?

not right now...just need to find some time.

tripodsan:
@wildone https://github.com/apache/jackrabbit/pull/88 is stil not fixed...until then, we cannot proceed with this.

wildone:
@tripodsan thank you, will follow up.

kwin:
Obsoleted by #92.

PR #68, opened 2019-11-29
JCRVLT-391 get rid of embedded Xerces

kwin:
Still WIP, some tests are failing


kwin:
Being obsoleted by #69.

PR #69, opened 2019-12-01
JCRVLT-391 replace embedded xerces by StAX

kwin:


kwin:
Fixed in http://svn.apache.org/viewvc?view=revision&revision=1871734.

PR #72, opened 2020-03-03
JCRVLT-415 enforce correct primary type for nodes being created by

kwin:
FolderArtifactHandler


kwin:
@tripodsan Can you have a look?

vault-core/src/main/java/org/apache/jackrabbit/vault/fs/impl/io/FolderArtifactHandler.java
```diff
@@ -92,6 +92,9 @@ public ImportInfoImpl accept(WorkspaceFilter wspFilter, Node parent, String name
             }
 
             Node node = parent.getNode(dir.getRelativePath());
+            if (wspFilter.contains(node.getPath()) && !node.getPrimaryNodeType().isNodeType(nodeType)) {
```
tripodsan: afir, `isNodeType()` also allows the node to be of a sub-type of the given node type. eg: if the node is `sling:Folder` and you check `isNodeType('nt:hierachyNode')` it would return `true`....

I think here you want to check for equality?

kwin: Good catch, fixed with https://github.com/apache/jackrabbit-filevault/pull/72/commits/440c535d103ce73733bc6a22bc13f51f1869fba7.

PR #73, opened 2020-03-06
JCRVLT-417 create intermediate nodes with default type of parent

kwin:
(instead of nt:folder)


kwin:
The problem is that "nt:folder" itself does not define a default child node type, so we need a fallback for that.

PR #79, opened 2020-05-19
Bugfix/jcrvlt 436

kwin:


kwin:
Merged in http://svn.apache.org/r1877947.

PR #80, opened 2020-05-25
JCRVLT-427 configurable authorizable ids for executing hooks

kwin:


vault-core/src/main/java/org/apache/jackrabbit/vault/packaging/registry/impl/AbstractPackageRegistry.java
```diff
@@ -59,6 +65,17 @@
      */
     public static final String DEFAULT_PACKAGE_ROOT_PATH_PREFIX = DEFAULT_PACKAGE_ROOT_PATH + "/";
 
+    protected final String[] additionalAuthorizableIdsAllowedToExecuteHooks;
+
+    protected final String[] additionalAuthorizableIdsAllowedToInstallPackagesRequiringRoot;
```
tripodsan, line 70: I think those a very long field names :-)

vault-core/src/main/java/org/apache/jackrabbit/vault/packaging/impl/JcrPackageManagerImpl.java
```diff
@@ -88,8 +88,8 @@
      * @param session repository session
      * @param roots the root paths to store the packages.
      */
-    public JcrPackageManagerImpl(Session session, String[] roots) {
-        this(new JcrPackageRegistry(session, roots));
+    public JcrPackageManagerImpl(@NotNull Session session, @Nullable String[] roots, @Nullable String[] additionalAuthorizableIdsAllowedToExecuteHooks, @Nullable String[] additionalAuthorizableIdsAllowedToInstallPackagesRequiringRoot) {
+        this(new JcrPackageRegistry(session, additionalAuthorizableIdsAllowedToExecuteHooks, additionalAuthorizableIdsAllowedToInstallPackagesRequiringRoot, roots));
```
tripodsan, line 92: I don't like this in particular:

- what about creating a `JcrPackageRegistryConfig` that holds the `roots`, `authIdsExec` and `authIdsRootInstall` value?
- or at least creating a `JcrPackageRegistrySecurity` config that holds the `authIdsExec` and `authIdsRootInstall` values?

in any case: also add default constructors that implement the old api.

vault-core/src/main/java/org/apache/jackrabbit/vault/packaging/registry/impl/AbstractPackageRegistry.java
kwin: Indeed, but I like verbosity over short names. Do you have any other proposal?

vault-core/src/main/java/org/apache/jackrabbit/vault/packaging/impl/JcrPackageManagerImpl.java
kwin: For me this is not API, as this is in a not exported impl package, so keeping backwards-compatibility is not necessary as this can only be used by the bundle itself. But wrapping the config makes sense.

tripodsan: it's not about backward compatibility, but to write `...null, null)` alot is kind of disturbing.

kwin:
@tripodsan Thanks for your feedback. Can you check again now? I restored the old constructors and added a wrapper for the options.

vault-core/src/main/java/org/apache/jackrabbit/vault/packaging/registry/impl/AbstractPackageRegistry.java
```diff
@@ -44,6 +50,24 @@
  */
 public abstract class AbstractPackageRegistry implements PackageRegistry, InternalPackageRegistry {
 
+    public static final class SecurityConfig {
+        private final String[] authIdsForHookExecution;
+        private final String[] authIdsForRootInstallation;
+        
+        public SecurityConfig(String[] authIdsForHooks, String[] authIdsForRoots) {
+            super();
```
tripodsan, line 58: not needed.

vault-core/src/main/java/org/apache/jackrabbit/vault/packaging/registry/impl/AbstractPackageRegistry.java
```diff
@@ -59,6 +83,18 @@
      */
     public static final String DEFAULT_PACKAGE_ROOT_PATH_PREFIX = DEFAULT_PACKAGE_ROOT_PATH + "/";
 
+    protected final @NotNull SecurityConfig securityConfig;
+
+    public AbstractPackageRegistry(SecurityConfig securityConfig) {
+        super();
```
tripodsan, line 89: not needed.

vault-core/src/main/java/org/apache/jackrabbit/vault/packaging/registry/impl/FSPackageRegistry.java
```diff
@@ -138,15 +138,31 @@ public FSPackageRegistry(@NotNull File homeDir) throws IOException {
...
         this.scope = scope;
         loadPackageCache();
     }
-
     /**
      * Deafult constructor for OSGi initialization (homeDir defined via activator)
+     * @throws IOException 
      */
-    public FSPackageRegistry() {
+    @Activate
```
tripodsan, line 155: since when can you use a constructor as activator?

kwin: Since DS 1.4 (supported since AEM 6.3): https://docs.osgi.org/specification/osgi.cmpn/7.0.0/service.component.html#service.component-constructor.injection

PR #82, opened 2020-05-25
JCRVLT-438 add RCP API

kwin:


vault-rcp/src/main/java/org/apache/jackrabbit/vault/rcp/impl/RcpTaskImpl.java
```diff
@@ -211,18 +210,22 @@ public STATE getState() {
...
     }
 
+    @Override
     public String getDestination() {
         return dst;
     }
 
+    @Override
     public void setRecursive(boolean b) {
         this.recursive = b;
```
tripodsan, line 225: don't allow modification if the task is running.

vault-rcp/src/main/java/org/apache/jackrabbit/vault/rcp/impl/RcpTaskImpl.java
```diff
@@ -211,18 +210,22 @@ public STATE getState() {
...
     }
 
+    @Override
     public void addExclude(String exclude) throws ConfigurationException {
```
tripodsan, line 229: don't allow modification if the task is running.

vault-rcp/src/main/java/org/apache/jackrabbit/vault/rcp/impl/RcpTaskImpl.java
kwin: I will make recursive final

vault-rcp/src/main/java/org/apache/jackrabbit/vault/rcp/impl/RcpTaskImpl.java
kwin: I will make the filter final

PR #83, opened 2020-06-02
JCRVLT-398 - migrate to git

kwin:


kwin:
Regarding SCM elements: I would like to leverage http://maven.apache.org/ref/3.6.3/maven-model-builder/index.html#Inheritance_Assembly and also releases are only done on the full reactor

PR #89, opened 2020-06-26
JCRVLT-449 persist RCP tasks

kwin:
make WorkspaceFilter serializable


kwin:
It is java native serialization. nt:unstructured or json/xml nt:file would be more effort.

kwin:
@tripodsan Do you think it would be ok to depend on https://github.com/FasterXML/jackson-databind for JSON de/serialization?

tripodsan:
> @tripodsan Do you think it would be ok to depend on https://github.com/FasterXML/jackson-databind for JSON de/serialization?

I think so...or use XML.

I don't like the binary format, because it's not transparent to the user and not editable.

kwin:
Now I persist (partly) in the repo and use JSON format.
@tripodsan Please have a look again

vault-core/src/main/java/org/apache/jackrabbit/vault/fs/api/package-info.java
```diff
@@ -15,7 +15,7 @@
  * limitations under the License.
  */
 
-@Version("2.7.0")
+@Version("2.8.0")
```
tripodsan, line 18: why isn't this a patch increase?

vault-rcp/src/main/java/org/apache/jackrabbit/vault/rcp/impl/FileNodeBackedInputStream.java
```diff
@@ -0,0 +1,52 @@
...
+import org.apache.jackrabbit.commons.JcrUtils;
+import org.apache.sling.jcr.api.SlingRepository;
+
+public class FileNodeBackedInputStream extends InputStream {
+
+    private final Session session;
+    private final InputStream input;
+    
+    FileNodeBackedInputStream(SlingRepository repository, String nodePath) throws RepositoryException {
+        session = repository.loginService(null, null);
```
tripodsan, line 35: this is not good. it should run with a session provided by the user who initiates the serialization

vault-rcp/src/main/java/org/apache/jackrabbit/vault/rcp/impl/FileNodeBackedOutputStream.java
```diff
@@ -0,0 +1,68 @@
...
+
+public class FileNodeBackedOutputStream extends OutputStream {
+
+    private final Session session;
+    private final ByteArrayOutputStream output;
+    private final String nodePath;
+    private final String mimeType;
+
+    public FileNodeBackedOutputStream(SlingRepository repository, String nodePath, String mimeType) throws RepositoryException {
+        session = repository.loginService(null, null);
```
tripodsan, line 41: this is not good, it should run with a session tied to the user that runs the export.

kwin:
This is not about explicit import/export, but rather an implicit persist feature on all tasks. It would be possible to persist every time a new task is added and reuse that session, but there is no session nor credentials available at the time the service is started. Do you have any specific security concerns with the system session?

vault-core/src/main/java/org/apache/jackrabbit/vault/fs/api/package-info.java
kwin: Don't really get that. This is a new feature and shouldn't be backported at all.

tripodsan:
> This is not about explicit import/export, but rather an implicit persist feature on all tasks. It would be possible to persist every time a new task is added and reuse that session, but there is no session nor credentials available at the time the service is started. Do you have any specific security concerns with the system session?

I guess, the entire RCP service is kind of insecure and should require an admin session for adding a task.
Also, I think that just persisting the task isn't enough, as the current progress is not saved. i.e. after ever copy operation,
the current traverse state should be stored, otherwise this feature is useless.

Also, I think I changed my mind, and I think it's better to store the task in the bundle data than in the repository.

kwin:
> I guess, the entire RCP service is kind of insecure and should require an admin session for adding a task.

I don't follow here, because starting a task via API requires giving a session on the destination system: https://github.com/apache/jackrabbit-filevault/blob/70dfb76e5c5aef46866b6e31570ce6cea9c9ccd7/vault-rcp/src/main/java/org/apache/jackrabbit/vault/rcp/impl/RcpTaskImpl.java#L195. The servlet is using the current session for that (i.e. using the context of the calling user).

> Also, I think that just persisting the task isn't enough, as the current progress is not saved. i.e. after ever copy operation,
the current traverse state should be stored, otherwise this feature is useless.

For me the persist feature is mostly about backstaging content from PROD to STAGE or other environments which happens periodically. For one time migration things it is not useful. Therefore I would also not persist the state, but rather start tasks from scratch manually after a restart

> Also, I think I changed my mind, and I think it's better to store the task in the bundle data than in the repository.

Why is that? Having it in the repo allows to configure tasks via packages instead of via Servlet/API only. I don't see any security implications as starting tasks still require a dedicated session!

tripodsan:
> Why is that? Having it in the repo allows to configure tasks via packages instead of via Servlet/API only. I don't see any security implications as starting tasks still require a dedicated session!

ok..makes sense. so you would store it in the 'stopped' state without any credentials and progress information. then I guess the repository is ok.

tripodsan:
but still, writing the task to the repository should use the user session.
for reading, it is ok to use the system session.

btw: how to you specify the location? maybe it would be better to store it as bundle configuration and then it gets automatically saved to /apps/config... no?

kwin:
> but still, writing the task to the repository should use the user session.
for reading, it is ok to use the system session.

Would require API changes, as currently only starting requires passing a session. Also I don't see the additional benefit here.

> btw: how to you specify the location? maybe it would be better to store it as bundle configuration and then it gets automatically saved to /apps/config... no?

I am already using a configurable location: https://github.com/apache/jackrabbit-filevault/pull/89/files#diff-c8d17df3d0f7629d1f12670077416b5dR76
Storing it as OSGi config will not work with composite nodestores as /apps is readonly then

tripodsan:
> Would require API changes, as currently only starting requires passing a session. Also I don't see the additional benefit here.
> 

An attacker could add tasks that copy (confidential) content to his own repository and then _somehow_ manage to get the start started. especially if the entire process is automated, this could easily happen unnoticed.

kwin:
@tripodsan Now I use only OSGi configuration. That way one can deploy a task configuration via regular OSGi config deployment means (e.g. https://sling.apache.org/documentation/bundles/osgi-installer.html). Still no one can easily read those files without having access to the Felix Web Console or the file system. WDYT?

tripodsan:
> @tripodsan Now I use only OSGi configuration. That way one can deploy a task configuration via regular OSGi config deployment means (e.g. https://sling.apache.org/documentation/bundles/osgi-installer.html). Still no one can easily read those files without having access to the Felix Web Console or the file system. WDYT?

that's great! maybe we tackle the authentication of the write operation in the servlet in a separat PR ? I really think, one should be admin to do this.

kwin:
> An attacker could add tasks that copy (confidential) content to his own repository and then somehow manage to get the task started. especially if the entire process is automated, this could easily happen unnoticed.

IMHO the target is always the local repo, i.e. it is always a pull (never a push) as the session being used in https://github.com/apache/jackrabbit-filevault/blob/c0d35641aa761a1109137a82dc301bf768dea0c4/vault-rcp/src/main/java/org/apache/jackrabbit/vault/rcp/impl/RcpTaskImpl.java#L239 is always a local session (if started via the servlet). Only the source session is a remote one (https://github.com/apache/jackrabbit-filevault/blob/c0d35641aa761a1109137a82dc301bf768dea0c4/vault-rcp/src/main/java/org/apache/jackrabbit/vault/rcp/impl/RcpTaskImpl.java#L211). This is different than the CLI RCP command!

tripodsan:
> > This special vault rcp version can only be used to import content from remote repositories.

oh...i forgot :-)

PR #91, opened 2020-07-02
Added NodeExporter to export only the content of a primary artifact when a flag is set in the ExportOptions

reschke:


vault-core/src/main/java/org/apache/jackrabbit/vault/fs/io/NodeExporter.java
```diff
@@ -0,0 +1,84 @@
...
+
+import javax.jcr.Node;
+import javax.jcr.RepositoryException;
+
+import org.apache.commons.io.IOUtils;
+import org.apache.jackrabbit.vault.fs.api.Artifact;
+import org.apache.jackrabbit.vault.fs.api.ArtifactType;
+import org.apache.jackrabbit.vault.fs.api.VaultFile;
+
+public class NodeExporter extends AbstractExporter {
```
kwin, line 32: This seems on the wrong API level, a node is not really a package. Can you work with the `DocViewSAXFormatter` and `DocViewSAXImporter` directly? Maybe we should just export that as API...

PR #76, opened 2020-05-04
JCRVLT-419: Feature/enable insecure https host

tripodsan:
same as #64 but with fixed imports

PR #92, opened 2020-07-08
JCRVLT-419 allow to parameterize connection options

kwin:
Depends upon https://github.com/apache/jackrabbit/pull/92.

obsoletes both #76 and #64.


kwin:
@tripodsan WDYT about this proposal? I would like to do a release soon (tm)....

PR #100, opened 2020-10-20
JCRVLT-433 access control validator

kwin:
validates that achandling != (ignore|clear) in case an acl is found in
the package


tripodsan (changes requested):
nice one.

tripodsan (approved):
oops.. wrong button

PR #104, opened 2020-11-26
JCRVLT-479 do not validate auto-created properties

stefanseifert:


kwin:
LGTM, but I would like to have a test.

stefanseifert:
unit test added

PR #108, opened 2020-12-03
JCRVLT-485 unit test to reproduce the problem with mix:versionable properties [WIP]

stefanseifert:
this is a draft PR to create a unit test that reproduces the problem (which it does not yet).


kwin:
There was a call to `Validator.done()` missing. I added that in https://github.com/apache/jackrabbit-filevault/commit/6be056783fff2db115c03d91409965589101a2a4 and merged to master.

PR #109, opened 2020-12-03
JCRVLT-486 introduce new configuration property "validNameSpaces"

stefanseifert:
allowing to configure list of allowed namespaces in the format: prefix1=ns-uri1,prefix2=nsuri2,...


kwin:
LGTM, but can you document that also at https://github.com/apache/jackrabbit-filevault/blob/master/vault-doc/src/site/markdown/validation.md?

stefanseifert:
update PR with documentation and an additional unit test

PR #110, opened 2020-12-07
JCRVLT-485 Ignore mandatory protected properties from mix:versionable

stefanseifert:


vault-validation/src/main/java/org/apache/jackrabbit/vault/validation/spi/impl/nodetype/JcrNodeTypeMetaDataImpl.java
```diff
@@ -89,6 +89,9 @@
 
     private static final Collection<Name> ALLOWED_PROTECTED_PROPERTIES = Arrays.asList(NameConstants.JCR_PRIMARYTYPE,
             NameConstants.JCR_MIXINTYPES);
+    // ignore mandatory protected properties from mix:versionable, they are handled specifically in filevault (JCRVLT-485)
```
kwin, line 92: Probably we should exclude all properties listed in https://github.com/apache/jackrabbit-filevault/blob/f785fcb24d4cbd01c734e9273310a925c29ae15b/vault-core/src/main/java/org/apache/jackrabbit/vault/fs/impl/io/DocViewSAXImporter.java#L921 from both the mandatory check as well as the protected check. WDYT?

stefanseifert: yes, i think this makes sense - updated the PR

kwin (approved):

kwin:
I committed a slightly modified version in https://github.com/apache/jackrabbit-filevault/commit/0fa305d5b0f6dec43e181db3fabeb32603d69501.

PR #114, opened 2021-01-07
Fix filter mode attribute definition in filter.xsd

karollewandowski:
I noticed that the filter `mode` attribute is simply defined as a string, rather than a reference to the `mode` attribute definition from line 33.
It may result in incorrect schema validations, because any string will be allowed, instead of: `replace`, `merge`, `update`.


kwin:
@karollewandowski Thanks for the contribution, it is merged and part of the upcoming release.

PR #116, opened 2021-01-18
JCRVLT-493 document package types

kwin:


kwin:
@DominikSuess would appreciate your feedback as well.

DominikSuess (approved):
Looks pretty good to me - esp linking to
the real life applications of the Type usage  will certainly help to understand the context, reasoning and the purpose of corresponding checks.

tripodsan (approved):

PR #126, opened 2021-02-24
JCRVLT-448 always close output stream passed to

kwin:
JcrPackageManager.rewrap(...) and assemble(...)


tripodsan (changes requested):
I'm not comfortable with this change. it might have side effects for clients, expecting the output stream to be still open. eg when streaming a http response.

if you really need this, I'd prefer to add something like:

```
JarExporter withCloseStream(boolean value) {
  this.closeInputStream = value;
  return this
}
```

so you can use: 

```
JarExporter exp = new JarExporter(out).withCloseStream(true)
```

kwin:
>  it might have side effects for clients, expecting the output stream to be still open. eg when streaming a http response.

The JarExporter already closes the underlying output stream in 90% of the cases. I just made it consistent so that this even works, when `open()` has not been called.

tripodsan (approved):
ok. thanks.

PR #128, opened 2021-03-11
JCRVLT-511 add package task options

kwin:


vault-core/src/main/java/org/apache/jackrabbit/vault/packaging/registry/PackageTaskBuilder.java
```diff
@@ -37,9 +37,17 @@
     /**
      * Sets the type of this task
      * @param type the type
-     * @return this.
+     * @return the parent execution plan builder.
```
kwin, line 40: @tripod Was that really supposed to return the `ExecutionPlanBuilder` instead of `PackageTaskBuilder`?

PR #129, opened 2021-03-16
JCRVLT-506 remove handling of JCR 1.0 nt:nodeType

kwin:


kwin:
I would tend to remove `NodeTypeAggregator` and `CNDSerializer` without replacement as I have never seen nodes of type `nt:nodeType` in real life. All repository implementation maintain node types only internally and only support serialization/deserialization to/from CND. WDYT?

tripodsan (approved):

PR #136, opened 2021-04-28
JCRVLT-517 fix package cache init in OSGi containers

kwin:
get rid of lazy-loading of package cache in packages()
improve test coverage
fix another issue in FSInstallState not correctly persisting excludes


tripodsan (commented):
@kwin I'm not comfortable with this change, I think we added the lazy loading for a reason. so that bundle activation is fast.....

kwin:
Any reason why lazy loading was not used in the other constructor?

tripodsan:
> Any reason why lazy loading was not used in the other constructor?

which other constructor?

kwin:
> which other constructor?

https://github.com/apache/jackrabbit-filevault/blob/7eb85af4e8a1072c1c0a72f90c59d67310d7b102/vault-core/src/main/java/org/apache/jackrabbit/vault/packaging/registry/impl/FSPackageRegistry.java#L166

Also lazy loading is not at all implemented correctly. It needs to be done IMHO on almost all operations (not only on `packages` and `contains` but also all `register` method which should throw exceptions in case of duplicate package ids). Lazy loading also requires proper synchronization, as multiple methods may be called from different threads. Also bundle activation is not deferred by long-running component activators. 

Why do you think that long-running activate() methods in DS are a problem?

tripodsan (approved):
ok then :-)

stoerr:
I did consider getting rid of lazy loading completely when doing #135 , but I'm afraid the original author might have had a good reason to do it. The problem is that the method loadPackages() is a quite expensive method, as it accesses the filesystem and opens all files to collect metadata. If the filesystem is a network filesystem (which might even make sense for distributing packages across instances), then this can even block indefinitely. That's fine in tests (at least within Jackrabbit Vault the FSPackageRegistryIT and your new test seem to be the only places the additional constructors are used), but that's probably not OK during OSGI activation. That might block your bundles from coming up and in the worst case you couldn't even get into the console to switch it off / change the path. And the registry might even not be used at all later...

So I'd strongly suggest to keep lazy loading for the activate method. You could consider encapsulating stateCache to switch it completely to lazy loading to be consistent (in the original code getInstallState / setInstallState methods also didn't call the loadPackages()), but that does change the behavior of the public constructors somewhat, so it might theoretically break e.g. a test someone did in a project using Vault, though that's admittedly a strange case. Not sure whether it's worth the risk.

vault-core/src/main/java/org/apache/jackrabbit/vault/packaging/registry/impl/FSPackageRegistry.java
```diff
@@ -185,6 +182,7 @@ public void activate(BundleContext context, Config config) {
         log.info("Jackrabbit Filevault FS Package Registry initialized with home location {}", this.homeDir.getPath());
         this.scope = InstallationScope.valueOf(config.scope());
         this.securityConfig = new AbstractPackageRegistry.SecurityConfig(config.authIdsForHookExecution(), config.authIdsForRootInstallation());
+        loadPackageCache();
```
stoerr, line 185: I think doing file system accesses within the OSGI activation is dangerous - please see my comment https://github.com/apache/jackrabbit-filevault/pull/136#issuecomment-828974792 .

vault-core/src/main/java/org/apache/jackrabbit/vault/packaging/registry/impl/FSInstallState.java
```diff
@@ -417,4 +422,84 @@ public WorkspaceFilter getFilter() {
     public Properties getProperties() {
         return properties;
     }
+
+    @Override
+    public int hashCode() {
+        final int prime = 31;
```
stoerr, line 428: Just for fun: I always thought using 31 as a factor is a quite bad choice, as it easily creates conflicts. https://stackoverflow.com/a/2816747/21499 explains why I think 92821 is much better when you are using that pattern. :-)

kwin:
I just checked again and the default timeout in Felix SCR for activate methods is 5 seconds (https://github.com/apache/felix-dev/blob/3e5671ae7e5107f4f849ef9d5f0a89b1ba9d7439/scr/src/main/java/org/apache/felix/scr/impl/manager/ScrConfiguration.java#L60), there are other alternatives outlined at https://stackoverflow.com/a/57250631.

vault-core/src/main/java/org/apache/jackrabbit/vault/packaging/registry/impl/FSInstallState.java
```diff
@@ -417,4 +422,84 @@ public WorkspaceFilter getFilter() {
...
+            if (other.subPackages != null)
+                return false;
+        } else if (!subPackages.equals(other.subPackages))
+            return false;
+        return true;
+    }
+
+    @Override
+    public String toString() {
+        return "FSInstallState [" + (packageId != null ? "packageId=" + packageId + ", " : "")
```
stoerr, line 498: Yes! I like it when you're adding toString() wherever it makes sense. That's often more pleasant in the debugger.

stoerr:
While we're at it: there are some log messages in FSPackageRegistry that could log the file / package they apply to to make it easier to find which of the potentially gazillion packages they apply to. I don't find a way to add a comment to these lines here in the PR since there are no changes near those, but I would suggest extending the log messages at lines 471, 581 (also typo in NOTREGISTERD), 684, 695, 704, 727 with some information about the package concerned. Thank you so much!

vault-core/src/main/java/org/apache/jackrabbit/vault/packaging/registry/impl/FSPackageRegistry.java
DominikSuess: The reason for not loading the cache here is because package operations are no frequent use case while the activate happens at any start of the system. This was designed for a Cloud-Service scenario with frequent starts & stops. Any expensive operation should be cut out of the startup procedure. The non-osgi initialization on the other hand is only required for initialization of an 'immutable' state of this is registry which would then only be read at system runtime (reused state) to validate dependencies for the mutable part being handled by the jcr registry. In case we have new scenarios we should make sure to not sacrifice those performance aspects.

DominikSuess (changes requested):
Please don't introduce costly synchronous processing in the osgi startup as package activities are rather a rare scenario in contrast to system start-up in containerized Cloud-Service setups.

vault-core/src/main/java/org/apache/jackrabbit/vault/packaging/registry/impl/FSPackageRegistry.java
DominikSuess: Clarification - installation of packages happens with an osgi startup at the moment as other mechanisms may write to repositorylike repo init but the accessible containers would only initialize the cache when there is package installation activity which is a rare scenario and is not supposed to delay startup

tripodsan (changes requested):
given the comments from the others, please remove the package loading from the activation method.

vault-core/src/main/java/org/apache/jackrabbit/vault/packaging/registry/impl/FSPackageRegistry.java
kwin: Loading the package cache is crucial before running ExecutionPlanBuilder (look at https://github.com/apache/sling-org-apache-sling-jcr-packageinit/pull/7/files). But I will come up with a fix which keeps prevents reading the state from FileSystem in case the FSPackageRegistry is just referenced but no methods are called (as it is the case sometimes in https://github.com/apache/sling-org-apache-sling-jcr-packageinit/blob/master/src/main/java/org/apache/sling/jcr/packageinit/impl/ExecutionPlanRepoInitializer.java

tripodsan (approved):
I can't really judge the changes... but I like the decoupling of the install state.

PR #137, opened 2021-05-07
Bugfix/jcrvlt 255 importmodes for docview

kwin:


kwin:
@tripodsan Can you have a look? This is a delicate topic but hopefully the new ImportModes make FileVault behave much more consistent...

vault-core/src/main/java/org/apache/jackrabbit/vault/fs/impl/io/FileArtifactHandler.java
```diff
@@ -257,7 +250,9 @@ public ImportInfoImpl accept(WorkspaceFilter wspFilter, Node parent,
                     path = path.substring(idx + 1);
                 }
                 // only update binary if import mode is not MERGE
-                if (wspFilter.getImportMode(parentNode.getPath()) != ImportMode.MERGE) {
+                // TODO: why? if not existing yet even with merge should be set afterwards
```
kwin, line 253: @tripodsan Do you know why MERGE should not even set the property if it is not there yet?

tripodsan: well, merge must not modify the property.  since the binary property is required on the file, it doesn't need to be processed here (i think :-)

tripodsan (commented):
as you said, it's quite a delicate issue..and I wouldn't touch it. the risk breaking existing workflows is too high :-)

kwin:
There is a desperate need for having proper merge import modes which work on all node structures. I therefore added new ImportModes. The idea is that the old ones work exactly as before. Do you see any places where this PR broke backwards-compatibility?

tripodsan:
> Do you see any places where this PR broke backwards-compatibility?

I can't judge :-) maybe @DominikSuess or @reschke  can run it through the AEM smoke tests?

kwin:
TBH: Every relevant test for content packages should rather be in this repo and not somewhere else...
I haven't found much ITs for the ImportMode behaviour unfortunately so every increase in test coverage here is highly appreciated...

vault-core/src/main/java/org/apache/jackrabbit/vault/fs/impl/io/FileArtifactHandler.java
```diff
@@ -218,29 +220,20 @@ public ImportInfoImpl accept(WorkspaceFilter wspFilter, Node parent,
...
-                            info.merge(handler.getInfo());
-                        } catch (ParserConfigurationException e) {
-                            throw new RepositoryException(e);
-                        } catch (SAXException e) {
-                            throw new RepositoryException(e);
-                        }
-                    } else {
-                        info.onNop(path);
+                    try {
+                        DocViewSAXImporter handler = new DocViewSAXImporter(newParent, newName, newSet, wspFilter);
```
kwin, line 224: @tripodsan I couldn't construct a package where this handler was ever used for docview xml. Do you have an example package structure. Although the comment above mentions "full coverage files below a intermediate node" I couldn't manage to create an Artifact `SerializationType.XML_DOCVIEW` which was not `ArtifactType.PRIMARY`. Also for intermediate handling in https://github.com/apache/jackrabbit-filevault/blob/a3ea981d88a3004bcbc05080ac4f44208696b087/vault-core/src/main/java/org/apache/jackrabbit/vault/fs/io/Importer.java#L932 this handler is never used.

tripodsan (approved):

kwin:
@tripodsan What about the AEM smoke tests? Have you executed them against this PR?

tripodsan:
> @tripodsan What about the AEM smoke tests? Have you executed them against this PR?

no... I don't work with AEM anymore, hence I asked @reschke and @DominikSuess

reschke:
> no... I don't work with AEM anymore, hence I asked @reschke and @DominikSuess

Not working on AEM either.

PR #139, opened 2021-05-12
JCRVLT-521 drop support of Jackrabbit 2.14, Oak 1.6 and Declarative

kwin:
Services 1.3


tripodsan (commented):
I cannot judge the PR. @DominikSuess can you review ?

parent/pom.xml
```diff
@@ -517,12 +517,12 @@ Bundle-Category: jackrabbit
             <dependency>
                 <groupId>org.apache.commons</groupId>
                 <artifactId>commons-lang3</artifactId>
-                <version>3.9</version>
+                <version>3.6</version>
```
reschke, line 520: why the downgrade?

parent/pom.xml
```diff
@@ -532,12 +532,12 @@ Bundle-Category: jackrabbit
             <dependency>
                 <groupId>org.apache.httpcomponents</groupId>
                 <artifactId>httpmime</artifactId>
-                <version>4.5.3</version>
+                <version>4.5.8</version>
```
reschke, line 535: These should use the most recent version (4.5.13)

parent/pom.xml
kwin: In order to be still compatible with AEM 6.4. For OSGi bundles you rather reference the minimum version you are compatible with.

parent/pom.xml
kwin: Same as above

PR #142, opened 2021-05-31
JCRVLT-525 fix corruption of the XML files of the FSPackageRegistry

stoerr:
When the FSPackageRegistry is updated such that the XML files that save the state for each package will become corrupt if they are shorter than before. They are not truncated when they are rewritten, and if they become shorter, there are remnants of the old XML after the end of the newly written XML, so that there is a SAXParseException when they are read again). This is a simple fix that solves that problem and makes it usable again.

### DIFF
--- a/vault-core/src/main/java/org/apache/jackrabbit/vault/packaging/registry/impl/FSInstallState.java
+++ b/vault-core/src/main/java/org/apache/jackrabbit/vault/packaging/registry/impl/FSInstallState.java
@@ -302,7 +302,7 @@ public class FSInstallState {
      */
     public void save(Path file) throws IOException {
         Files.createDirectories(file.getParent());
-        try (OutputStream out = Files.newOutputStream(file, StandardOpenOption.CREATE)) {
+        try (OutputStream out = Files.newOutputStream(file)) {
             save(out);
         }
     }


vault-core/src/main/java/org/apache/jackrabbit/vault/packaging/registry/impl/FSInstallState.java
```diff
@@ -302,7 +302,7 @@ private static WorkspaceFilter readWorkspaceFilter(Element child) throws Configu
      */
     public void save(Path file) throws IOException {
         Files.createDirectories(file.getParent());
-        try (OutputStream out = Files.newOutputStream(file, StandardOpenOption.CREATE)) {
+        try (OutputStream out = Files.newOutputStream(file, StandardOpenOption.CREATE, StandardOpenOption.TRUNCATE_EXISTING)) {
```
kwin, line 305: I would rather rely on the standard options of https://docs.oracle.com/javase/7/docs/api/java/nio/file/Files.html#newOutputStream(java.nio.file.Path,%20java.nio.file.OpenOption...)

> If no options are present then this method works as if the CREATE, TRUNCATE_EXISTING, and WRITE options are present

stoerr: Ah, indeed, thanks. Changed that. I'm wondering whether the alternative approach https://github.com/stoerr/jackrabbit-filevault/commit/916c7d0238435d8843ea506c95a87e70dc2bc6c8 would be a little better, though.

PR #143, opened 2021-06-04
JCRVLT-526 PackageProperties timezone format extension

stoerr:
When packages e.g. a creation date with a timezone format +- hhmm or +-hh are entered into the FSPackageRegistry, the date cannot be read by org.apache.jackrabbit.vault.packaging.impl.PackagePropertiesImpl.getCreated() : it returns null since it internally uses org.apache.jackrabbit.util.ISO8601.parse to parse the date, which doesn't support that (compare https://issues.apache.org/jira/browse/JCR-4267). This is a problem, since com.day.jcr.vault:content-package-maven-plugin (I tried versions 0.0.12 and 0.5.4) write package properties using such a format - e.g. 2021-05-26T15:12:21.673+0200 , while org.apache.jackrabbit.util.ISO8601.parse just supports a format like 2021-05-26T15:12:21.673+02:00 . 

This extends the PackageProperties to deal with these.


vault-core/src/main/java/org/apache/jackrabbit/vault/packaging/impl/PackagePropertiesImpl.java
```diff
@@ -211,22 +216,34 @@ public boolean requiresRestart() {
...
         try {
             // TODO: add timezone if not there?
             String p = getProperty(name);
-            return p == null
-                    ? null
-                    : ISO8601.parse(p);
+            if (p != null) {
+                result = ISO8601.parse(p);
+                if (result == null) {
+                    // Perhaps this is due to unusual timezone format +02 or +02:00 that isn't supported by ISO8601.parse
```
kwin, line 232: the comment is confusing, isn't the pattern only supporting +02 or 0200 (without colon)?

vault-core/src/main/java/org/apache/jackrabbit/vault/packaging/impl/PackagePropertiesImpl.java
```diff
@@ -211,22 +216,34 @@ public boolean requiresRestart() {
...
+                    // but sometimes produced by maven plugins, compare JCRVLT-526
```
kwin, line 233: the more suitable reference is IMHO https://issues.apache.org/jira/browse/JCRVLT-276

kwin (changes requested):
Thanks for the patch. Looks good, only minor nitpicks...

vault-core/src/main/java/org/apache/jackrabbit/vault/packaging/impl/PackagePropertiesImpl.java
stoerr: Right, I confused that. Thanks!

vault-core/src/main/java/org/apache/jackrabbit/vault/packaging/impl/PackagePropertiesImpl.java
stoerr: Done.

vault-core/src/main/java/org/apache/jackrabbit/vault/packaging/impl/PackagePropertiesImpl.java
```diff
@@ -211,22 +216,34 @@ public boolean requiresRestart() {
...
             String p = getProperty(name);
-            return p == null
-                    ? null
-                    : ISO8601.parse(p);
+            if (p != null) {
+                result = ISO8601.parse(p);
+                if (result == null) {
+                    // Perhaps this is due to unusual timezone format +02 or +0200 that aren't supported by ISO8601.parse
+                    // but sometimes produced by maven plugins, compare https://issues.apache.org/jira/browse/JCRVLT-276
+                    Matcher splittedDate = TIMEZONE_FIX_PATTERN.matcher(p);
```
jsedding, line 234: Rather than fixing things here, why not use `GregorianCalendar.from(ZonedDateTime.parse(p, DateTimeFormatter.ISO_OFFSET_DATE_TIME))`? Whether you then still call `ISO8601.parse(p)` before or not is a matter of taste.

kwin: According to https://docs.oracle.com/javase/8/docs/api/java/time/format/DateTimeFormatter.html#ISO_OFFSET_DATE_TIME and  https://docs.oracle.com/javase/8/docs/api/java/time/ZoneOffset.html#getId-- only supports the following timezone patterns:

> Z - for UTC (ISO-8601)
> +hh:mm or -hh:mm - if the seconds are zero (ISO-8601)
> +hh:mm:ss or -hh:mm:ss - if the seconds are non-zero (not ISO-8601)

But not something like `+hhmm` or just `+hh` or am I missing something here?

kwin: But based on this suggestion I clean up the impl  in https://github.com/apache/jackrabbit-filevault/pull/144. @stoerr @jsedding Please have a look

PR #144, opened 2021-06-07
JCRVLT-526 improved support of legacy timezone designators "+hhmm"

kwin:
Based on a suggestion by Julian Sedding

### DIFF
--- a/vault-core/src/main/java/org/apache/jackrabbit/vault/packaging/impl/PackagePropertiesImpl.java
+++ b/vault-core/src/main/java/org/apache/jackrabbit/vault/packaging/impl/PackagePropertiesImpl.java
@@ -19,16 +19,17 @@ package org.apache.jackrabbit.vault.packaging.impl;
 
 import java.net.URI;
 import java.net.URISyntaxException;
+import java.time.ZonedDateTime;
+import java.time.format.DateTimeFormatter;
+import java.time.format.DateTimeParseException;
 import java.util.Calendar;
 import java.util.Collections;
 import java.util.Enumeration;
+import java.util.GregorianCalendar;
 import java.util.HashMap;
 import java.util.Map;
 import java.util.Properties;
-import java.util.regex.Matcher;
-import java.util.regex.Pattern;
 
-import org.apache.jackrabbit.util.ISO8601;
 import org.apache.jackrabbit.vault.fs.io.AccessControlHandling;
 import org.apache.jackrabbit.vault.packaging.Dependency;
 import org.apache.jackrabbit.vault.packaging.PackageId;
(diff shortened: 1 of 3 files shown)


vault-core/src/main/java/org/apache/jackrabbit/vault/packaging/impl/PackagePropertiesImpl.java
```diff
@@ -224,21 +222,17 @@ public boolean requiresRestart() {
...
-                        String reformattedDate = splittedDate.group("dateUptoTzHours") + ":" + tzminutes;
-                        result = ISO8601.parse(reformattedDate);
-                    }
+                ZonedDateTime zonedDateTime;
+                try {
+                    // supports date in format new SimpleDateFormat("yyyy-MM-dd'T'HH:mm:ss.SSSXXX")
+                    zonedDateTime = ZonedDateTime.parse(p, DateTimeFormatter.ISO_OFFSET_DATE_TIME);
+                } catch (DateTimeParseException e) {
+                    // support dates given out via new SimpleDateFormat("yyyy-MM-dd'T'HH:mm:ss.SSSZ"); as well (used in package-maven-plugin till version 1.0.3, compare with https://issues.apache.org/jira/browse/JCRVLT-276)
+                    zonedDateTime = ZonedDateTime.parse(p, DateTimeFormatter.ofPattern("yyyy-MM-dd'T'HH:mm:ss.SSSX"));
```
stoerr, line 233: That undoubtedly looks much cleaner than the pattern replacement approach. So, if that is backwards compatible to ISO8601 and does parse the +0200 timezone format variant, then that's completely fine with me. Mind you - the test I wrote did not intend to see whether it still parses all date formatting variants ISO8601.parse understands, if there are some, since it was still using ISO8601.parse so far. So you might want to check whether you need to extend the test somewhat.

vault-core/src/test/java/org/apache/jackrabbit/vault/packaging/impl/PackagePropertiesImplTest.java
```diff
@@ -93,9 +94,6 @@ public void testDateFormat() {
         checkDateParsing("2021-05-26T15:12:21.673+0200","2021-05-26T13:12:21.673Z");
         checkDateParsing("2021-05-26T15:12:21.673-0230","2021-05-26T17:42:21.673Z");
 
-        // missing timezone is treated as UTC. TODO is that right?
-        checkDateParsing("2021-05-26T15:12:21.673","2021-05-26T15:12:21.673Z");
-
```
stoerr, line 98: What happens now if the timezone is missing? I don't know whether anything produces that date format, but at least I'm a fan of specifying such edge cases in the test, even if the documentation doesn't really say anything about that. Not sure whether it's a problem if that behavior changes. That throws an exception now, right?

jsedding (approved):
@kwin Thanks! This looks a lot better than parsing/fixing patterns with regexp.

Should we create a constant for `DateTimeFormatter.ofPattern("yyyy-MM-dd'T'HH:mm:ss.SSSX")`? I don't know how costly it is to recreate, and afaik DTFs are thread-safe.

I wonder if a `DateTimeFormatter` could be created to correctly parse all the expected inputs, but I couldn't quickly find a configuration that passes the tests.

vault-core/src/test/java/org/apache/jackrabbit/vault/packaging/impl/PackagePropertiesImplTest.java
kwin: That date format was never produced and therefore is not supported (i.e. returns null)

stoerr: It didn't return null, as the test says. But never mind.

PR #152, opened 2021-06-25
JCRVLT-542 - adding checks for importmode of filterPathSet along with…

DominikSuess:
… corresponding IT


kwin:
@DominikSuess Thanks for the PR. IIUC this is a regression, so I wonder which commit broke this. IMHO there was never a check for import mode in that handler, so even before https://github.com/apache/jackrabbit-filevault/commit/5f3b1d363da7bcf3c285a0bdbb394254ca3e57d1#diff-8d660cbc935be10303ebfd498eb1574cbba07f3ae8e565f470762b2e6f97b512 it was probably behaving like that. Do you have further insights since when this do no longer work?

kwin:
Ok, regression identified in https://issues.apache.org/jira/browse/JCRVLT-542?focusedCommentId=17369527&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-17369527. I came up with a slightly different fix because even mode UPDATE should never modify the primary type in https://github.com/apache/jackrabbit-filevault/pull/153. @DominikSuess Please have a look!

kwin:
Superseded by #153

PR #154, opened 2021-06-29
JCRVLT-543 ignore protected properties during import

kwin:
dynamically determine if a property is protected
fail with an exception for all other constraint violations (in mode =
REPLACE)


DominikSuess:
@kwin overall looks cleaner handling wise - specifically on not using the static list on removal check.

PR #155, opened 2021-06-30
JCRVLT-544 never create intermediate nodes of type nt:base or

kwin:
nt:hierarchyNode

Both types have too many restrictions to be useful as intermediate node
types

### DIFF
--- a/vault-core/src/main/java/org/apache/jackrabbit/vault/fs/impl/io/FolderArtifactHandler.java
+++ b/vault-core/src/main/java/org/apache/jackrabbit/vault/fs/impl/io/FolderArtifactHandler.java
@@ -87,7 +87,7 @@ public class FolderArtifactHandler extends AbstractArtifactHandler {
 
     private @Nullable String getDefaultPrimaryChildNodeType(Node parent, String intermediateNodeName) throws RepositoryException {
         EffectiveNodeType effectiveNodeType = EffectiveNodeType.ofNode(parent);
-        NodeDefinition nodeDefinition = effectiveNodeType.getChildNodeDef(nd -> nd.getDefaultPrimaryType() != null, intermediateNodeName);
+        NodeDefinition nodeDefinition = effectiveNodeType.getApplicableChildNodeDefinition(nd -> nd.getDefaultPrimaryType() != null, intermediateNodeName);
         if (nodeDefinition != null) {
             return nodeDefinition.getDefaultPrimaryTypeName();
         }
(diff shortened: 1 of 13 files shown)


kwin:
@DominikSuess Can you please add some ITs to verify this fixes your issue? In my AEM Quickstart SDK I don't see a package named `cq-integration-searchpromote-content`....

vault-core/src/main/java/org/apache/jackrabbit/vault/fs/impl/io/FolderArtifactHandler.java
```diff
@@ -62,6 +70,18 @@ public void setNodeType(String nodeType) {
         this.nodeType = nodeType;
     }
 
+    private Node createIntermediateNode(Node parent, String intermediateNodeName) throws RepositoryException {
+        // preferably use default (=primary) node type for intermediate nodes
+        NodeType primaryNodeType = parent.getPrimaryNodeType();
+        final Node node;
+        if (!DISALLOWED_PRIMARY_NODE_TYPE_NAMES.contains(primaryNodeType.getName()) &&
```
DominikSuess, line 77: @kwin  - this doesn't solve the issue at hand - it is not the primaryType of the parent that is the problem but the default nodetype - therefore the default nodetype would need to be detected and compared instead (IIUC thats part of the NodeDefinitions that can be obtained via getChildNodeDefinitions())

DominikSuess:
@kwin  - I can provide a package isolated to the troublesome case (containing an equivalent CND & content structure) later today as I have some other high prio tasks to finish before leaving for PTO end of the week.

kwin:
You need to, because I don't understand the issue from the description then...

DominikSuess:
I'm a bit behind my tasks - I hope to complete by today but here the structure
Node /libs/foo with NodeType my:test with ChildeNodeDefinition definig default primary type as nt:base
Node /libs/foo/bar in package with out nodetype definition (formerly nt:folder but gets created as nt:base now)
Node /libs/foo/bar/failed  with any kind of nodetype - fails on commit as /libs/foo/bar can't take childNodes

jsedding:
@kwin I have created a branch (see commit 7ed67c4f) with a slightly refactored `EffectiveNodeType` util. Most notably, I changed the return types of some methods to `Optional` and removed the `Predicate` arguments on the methods. Consumers of the API can filter the resulting `Optional` with custom predicates instead.

Feel free to merge or ignore as it's mostly a matter of taste.

kwin:
@jsedding Thanks for the suggestion. I added some comments to the commit. The timing was just a bit unfortunate as I was still reworking some aspects of my branch. Now I should be done, so either you propose the changes again on top of my branch, or you wait until it is merged and then propose a PR.

jsedding:
@kwin sorry, I won't be able to work on the proposed changes in the near future. SO please go ahead with the merge.

PR #157, opened 2021-07-13
JCRVLT-546 only increase minor version of o.a.j.v.fs as the removed

kwin:
classes could have never been referenced from other bundles in an OSGi
environment (due to private super class)


kwin:
@DominikSuess Please have a look.

tripodsan (approved):

PR #159, opened 2021-08-23
JCRVLT-551 try to set UUID from package

kwin:
add additional import options to tweak UUID collision behavior

### DIFF
--- a/vault-core/src/main/java/org/apache/jackrabbit/vault/fs/impl/io/DocViewSAXImporter.java
+++ b/vault-core/src/main/java/org/apache/jackrabbit/vault/fs/impl/io/DocViewSAXImporter.java
@@ -88,6 +88,7 @@ import org.xml.sax.helpers.AttributesImpl;
 /**
  * Implements an importer that processes SAX events from a (modified) document
  * view. The behaviour for existing nodes works as follows:
+ * TODO: better description
  * <pre>
  *
  * - extended docview always includes SNS indexes
@@ -928,7 +929,7 @@ public class DocViewSAXImporter extends RejectingEntityDefaultHandler implements
         VersioningState vs = new VersioningState(stack, node);
         Node updatedNode = null;
         // try to set uuid via sysview import if it differs from existing one
-        if (ni.uuid != null && !node.getIdentifier().equals(ni.uuid)) {
+        if (ni.uuid != null && !node.getIdentifier().equals(ni.uuid) && !"rep:root".equals(ni.primary)) {
             NodeStash stash = new NodeStash(session, node.getPath());
             stash.stash();
             Node parent = node.getParent();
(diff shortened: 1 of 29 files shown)


vault-core/src/main/java/org/apache/jackrabbit/vault/fs/impl/io/DocViewSAXImporter.java
```diff
@@ -809,89 +813,47 @@ private void handleAuthorizable(Node node, DocViewNode ni) throws RepositoryExce
...
+                        // does uuid already exist in the repo?
+                        Node sameIdNode = session.getNodeByIdentifier(ni.uuid);
+                        // edge-case: same node path -> uuid is kept
+                        if (existingNode != null && existingNode.getPath().equals(sameIdNode.getPath())) {
+                            log.debug("Node with existing identifier is being updated without modifying its uuid");
+                        } else {
+                            // uuid found in path covered by filter
+                            if (isIncluded(sameIdNode, 0)) {
+                                log.warn("Node identifier {} for to-be imported node {} already taken by {}, trying to release it.", ni.uuid, currentNode.getPath() + '/' + ni.label, sameIdNode.getPath());
+                                removeReferences(sameIdNode);
```
tripodsan, line 836: why removing the references? if the existing node is at the wrong place, it should be moved but the references should remain.

tripodsan (approved):
this is rather a big change. AFAICS, the default behaviour is still the same as before.

vault-core/src/main/java/org/apache/jackrabbit/vault/fs/impl/io/DocViewSAXImporter.java
kwin: IMHO if a node with a conflicting id already exists in the repo at a different path, one should either
- not touch it (if not contained in filter rules)
- remove it and its references (if contained in the filter rules)
Just assuming that this node should be moved because its id is the same is very dangerous as conflicts may occur for totally unrelated nodes (although unlikely)

tripodsan: the most common use case, is probably a rename/move of a referenceable node due to restructuring of a folder hierarchy. removing and recreating the node at the new place should be avoided. also, the references to it should remain. this is the idea of MOVE_EXISTING.

kwin:
The default behaviour changed: https://github.com/apache/jackrabbit-filevault/pull/159/files#diff-805491a0bcebe271c6e12fa9b2a97e452effc7963d619203e2e2cc616d954639.
But IMHO in a way which is safer than before.

vault-core/src/main/java/org/apache/jackrabbit/vault/fs/impl/io/DocViewSAXImporter.java
kwin: The moved node may be referenced via id and by path at the same time. If the move operation is just silently reverted through package install the path references are broken!

kwin: In any case references are only removed if they are contained in the filter rules, so this does not destroy anything which would not be overwritten anyhow later in the package installation.

PR #161, opened 2021-08-27
JCRVLT-552  - check for matching property definition must match multivalue definitions for single values

kwin:


kwin:
@DominikSuess Any opinion before I go ahead and merge?

PR #163, opened 2021-08-30
JCRVLT-549 - node cannot be deleted if it's a multi-mandatory child node

ankitaagar:


vault-core/src/main/java/org/apache/jackrabbit/vault/fs/impl/io/DocViewSAXImporter.java
```diff
@@ -1215,13 +1215,13 @@ public void endElement(String uri, String localName, String qName) throws SAXExc
                             }
                         } else {
                             if (wspFilter.getImportMode(path) == ImportMode.REPLACE) {
-                                importInfo.onDeleted(path);
                                 // check if child is not protected
                                 if (child.getDefinition().isProtected()) {
                                     log.debug("Refuse to delete protected child node: {}", path);
-                                } else if (child.getDefinition().isMandatory()) {
+                                } else if (child.getDefinition().isMandatory() && numChildren <= JcrConstants.NUM_MANDATORY_CHILDREN) {
```
tripodsan, line 1221: I don't think that this requires a constant.
```suggestion
                                } else if (child.getDefinition().isMandatory() && numChildren <= 1) {
```

vault-core/src/main/java/org/apache/jackrabbit/vault/fs/impl/io/DocViewSAXImporter.java
```diff
@@ -1215,13 +1215,13 @@ public void endElement(String uri, String localName, String qName) throws SAXExc
                             }
                         } else {
                             if (wspFilter.getImportMode(path) == ImportMode.REPLACE) {
-                                importInfo.onDeleted(path);
```
tripodsan, line 1218: why did you remove this?

vault-core/src/main/java/org/apache/jackrabbit/vault/fs/impl/io/DocViewSAXImporter.java
tripodsan: however, I think this check is not correct. you need to count the number of child nodes that match the property definition. just having any other child node is not enough.

tripodsan (changes requested):
- wrong check (need to count the number of child nodes that match the same property definition)
- adding a test case would be good

kwin:
Please add a failing test which succeeds with this patch.

vault-core/src/main/java/org/apache/jackrabbit/vault/fs/impl/io/DocViewSAXImporter.java
```diff
@@ -1247,6 +1248,21 @@ public void endElement(String uri, String localName, String qName) throws SAXExc
         }
     }
 
+    private int getSiblWithReqPrimType(Node child) throws RepositoryException {
```
kwin, line 1251: This IMHO not only returns the number of siblings but includes child itself. Also the method does not care about required primary type. Why not rename to `getNumChildrenOfPrimaryNodeType(Node parent, String primaryNodeType)`

vault-core/src/main/java/org/apache/jackrabbit/vault/fs/impl/io/DocViewSAXImporter.java
```diff
@@ -1247,6 +1248,21 @@ public void endElement(String uri, String localName, String qName) throws SAXExc
...
+
+        int count=0;
+        Node parent = child.getParent();
+        NodeIterator iter = parent.getNodes();
+        while (iter.hasNext()) {
+            Node sibling = iter.nextNode();
+            if(child.getPrimaryNodeType().toString()
```
kwin, line 1258: what about extended primary node types (https://docs.adobe.com/content/docs/en/spec/jsr170/javadocs/jcr-2.0/javax/jcr/nodetype/NodeType.html#isNodeType(java.lang.String))?

vault-core/src/test/java/org/apache/jackrabbit/vault/packaging/integration/MandatoryNode.java
```diff
@@ -0,0 +1,59 @@
...
+import org.junit.Test;
+
+
+public class MandatoryNode extends IntegrationTestBase {
+ 
+    @Test
+    public void testSimpleFileAggregateOverwritesChildNodes()
+            throws RepositoryException, IOException, PackageException {
+
+        extractVaultPackage("/test-packages/wcm-rollout-config-1.zip");
```
kwin, line 38: Can you provide minimal test packages in exploded form (similar to https://github.com/apache/jackrabbit-filevault/tree/master/vault-core/src/test/resources/test-packages/test_nodetype_on_merge.zip)

kwin (changes requested):

vault-core/src/test/java/org/apache/jackrabbit/vault/packaging/integration/MandatoryNode.java
```diff
@@ -0,0 +1,59 @@
...
+import org.apache.jackrabbit.api.security.user.UserManager;
+import org.apache.jackrabbit.api.security.user.Authorizable;
+import org.apache.jackrabbit.vault.packaging.PackageException;
+import org.junit.Test;
+
+
+public class MandatoryNode extends IntegrationTestBase {
+ 
+    @Test
+    public void testSimpleFileAggregateOverwritesChildNodes()
```
kwin, line 35: This test IMHO succeeds even without any code changes. Please make sure it fails without your other changes!

vault-core/src/main/java/org/apache/jackrabbit/vault/fs/impl/io/DocViewSAXImporter.java
tripodsan: or:

```
getNumChildrenWithType(Node parent, String primaryNodeType)
```

and then check using `isNodeType()` against the required type (not the child type), since the child you are testing could be an super type of the one that is required by the CND.

vault-core/src/main/java/org/apache/jackrabbit/vault/util/JcrConstants.java
```diff
@@ -231,7 +231,12 @@
      * Pseudo property jcr:score used with query results
      */
     public static final String JCR_SCORE = "jcr:score";
-
+    /**
+     * One child node is mandatory.
+     * A mandatory item is one that, if
+     * its parent node exists, must also exist.
+     */
+    public static final int NUM_MANDATORY_CHILDREN = 1;
```
tripodsan, line 239: you don't need this.

tripodsan (commented):
also think of caching the checks. if a node has a lot of mandatory children, this gets rather expensive.

vault-core/src/main/java/org/apache/jackrabbit/vault/fs/impl/io/DocViewSAXImporter.java
tripodsan: actually, this more complicated, if same name siblings are allowed and stupidly complex CNDs....but let's ignore this for now :-)

vault-core/src/main/java/org/apache/jackrabbit/vault/fs/impl/io/DocViewSAXImporter.java
```diff
@@ -1215,10 +1215,11 @@ public void endElement(String uri, String localName, String qName) throws SAXExc
                             }
                         } else {
                             if (wspFilter.getImportMode(path) == ImportMode.REPLACE) {
+
                                 // check if child is not protected
                                 if (child.getDefinition().isProtected()) {
                                     log.debug("Refuse to delete protected child node: {}", path);
-                                } else if (child.getDefinition().isMandatory() && numChildren <= JcrConstants.NUM_MANDATORY_CHILDREN) {
+                                } else if (child.getDefinition().isMandatory() &&  getSiblWithReqPrimType(child) < 0) {
```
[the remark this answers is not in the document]
ankitaagar: Missed the = here. It's <=0. 
If there is a single mandatory node and that can be replaced with another mandatory node of the same type.

vault-core/src/test/java/org/apache/jackrabbit/vault/packaging/integration/MandatoryNode.java
ankitaagar: Without the changes, it fails.

vault-core/src/main/java/org/apache/jackrabbit/vault/util/JcrConstants.java
ankitaagar: Will remove

vault-core/src/main/java/org/apache/jackrabbit/vault/fs/impl/io/DocViewSAXImporter.java
```diff
@@ -1295,6 +1300,31 @@ public void endElement(String uri, String localName, String qName) throws SAXExc
...
+            Optional<NodeDefinition>  childDef = ent.getApplicableChildNodeDefinition(sibling.getName(), sibling.getPrimaryNodeType());
+            Boolean test = (def == childDef) || (def != null && def.equals(childDef));
+
+            /*This is not working
+            if(def.equals(childDef)) { 
+                count++; 
+            }*/
+            if(Arrays.equals(def.get().getRequiredPrimaryTypes(),childDef.get().getRequiredPrimaryTypes())) {
+                count++; 
+            }
```
tripodsan, line 1323: @kwin according to @ankitaagar, the `def.equals(childDef)` is never true. do you know why?

vault-core/src/main/java/org/apache/jackrabbit/vault/fs/impl/io/DocViewSAXImporter.java
```diff
@@ -1295,6 +1300,31 @@ public void endElement(String uri, String localName, String qName) throws SAXExc
         }
     }
 
+    private int getNumChildrenWithType(Node parent, Node child) throws RepositoryException {
+
+        int count=0;
+
```
tripodsan, line 1306: The idea is:
- get the `NodeDefinition` of the child node that is to be deleted
- loop through all childnodes and count the ones that have the same nodedef
- if the nodedef is mandatory and there are more than 1 nodes with this nodedef, we can delete it.

vault-core/src/main/java/org/apache/jackrabbit/vault/fs/impl/io/DocViewSAXImporter.java
kwin: Currently https://github.com/apache/jackrabbit-filevault/blob/master/vault-core/src/main/java/org/apache/jackrabbit/vault/util/EffectiveNodeType.java does not implement equals, hash and toString

kwin: Added those methods now in https://github.com/apache/jackrabbit-filevault/commit/0087391e3e2dbe0f70fab5dcb1a283c354c13c25

tripodsan: thanks. but @ankitaagar is comparing the `Optional<NodeDefinition>`, and not the effective node type.

kwin: NodeDefinition includes the node name (https://docs.adobe.com/content/docs/en/spec/jsr170/javadocs/jcr-2.0/javax/jcr/nodetype/ItemDefinition.html) and therefore cannot ever be equal for two different (sibling) nodes.

tripodsan: no. it's the name of the item definition, not the name of the child node. i.e. in this case it should be `*`
https://docs.adobe.com/content/docs/en/spec/jsr170/javadocs/jcr-2.0/javax/jcr/nodetype/ItemDefinition.html#getName()

tripodsan: the problem is rather, that the [ItemDefintion](https://github.com/apache/jackrabbit-oak/blob/c6ddcc55bee3de915459af01e91edad32d538f3d/oak-core/src/main/java/org/apache/jackrabbit/oak/plugins/nodetype/ItemDefinitionImpl.java#L42) doesn't implement equals :-(

@ankitaagar I think your must implement the equals yourself or we add it as helper function to `EffectiveNodeType`

kwin: Right, but equals() is not implemented in https://github.com/apache/jackrabbit-oak/blob/trunk/oak-core/src/main/java/org/apache/jackrabbit/oak/plugins/nodetype/ItemDefinitionImpl.java either (nor does it necessarily have to according to the JCR 2.0 spec). Feel free to raise a PR for Oak.

ankitaagar:
After the merge and release of jackrabbit-oak, I will update the oak.version in parent/pom.xml

kwin:
Please come up with a solution which works with older Oak versions! FileVault is often used in older Oak versions and the related PR is unlikely to be backported to older Oak versions.

ankitaagar:
@kwin On which version of jackrabbit-oak should it be fixed?

kwin:
FileVault 3.5.x is compatible with with Oak 1.8 and Jackrabbit 2.16 (compare with https://downloads.apache.org/jackrabbit/filevault/3.5.0/RELEASE-NOTES.txt). Please come up with a fix in FileVault only which does not rely on equals in `ItemDefinitionImpl`.

ankitaagar:
@kwin We can backport the jackrabbit-oak fix to its 1.8 version. 1.8 will be released in few days.

vault-core/src/main/java/org/apache/jackrabbit/vault/fs/impl/io/DocViewSAXImporter.java
```diff
@@ -1295,6 +1300,26 @@ public void endElement(String uri, String localName, String qName) throws SAXExc
         }
     }
 
+    private int getNumChildrenWithType(Node parent, Node child) throws RepositoryException {
```
thomasmueller, line 1303: If there are many entries, counting can be slow. If all you need is "getNumChildrenWithType(child.getParent(), child) <= 1" then it's not needed to count the entries. What about "hasChildWithType(...)"?

tripodsan: the problem is, that a node can only be deleted, if there are more than one child nodes with the respective type.
so this method needs to count. but as an optimization, the counting can be aborted when there are found 2.

ankitaagar: I can change that to two

reschke:
So, the options seem to be:

- rewrite the patch so that it does not rely on "equals()", or
- make sure that "equals()" works on all applicable versions of Jackrabbit (classic) as well.

If we care strictly about being able to use any JCR implementation, I think only the first option will work.

tripodsan:
> * rewrite the patch so that it does not rely on "equals()", or

+1

reschke:
See 

  https://github.com/apache/jackrabbit-filevault/pull/163/commits/8af655c6b7c17cbf8fc8129de0aef9226cb4af51

If there's concern that comparing the type names is not sufficient, we could delegate to CompactNodeTypeDefWriter from jackrabbit-jcr-commons (which would emulate Oak's "NodeTypeImpl.equals()")

vault-core/src/main/java/org/apache/jackrabbit/vault/fs/impl/io/DocViewSAXImporter.java
```diff
@@ -1300,27 +1299,29 @@ public void endElement(String uri, String localName, String qName) throws SAXExc
         }
     }
 
-    private int getNumChildrenWithType(Node parent, Node child) throws RepositoryException {
-
-        int count=0;
-       
-        NodeIterator iter = parent.getNodes();
+    private boolean hasSiblingWithSameType(Node parent, Node child) throws RepositoryException {
```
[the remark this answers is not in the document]
reschke: True enough.

vault-core/src/test/java/org/apache/jackrabbit/vault/packaging/integration/MandatoryNode.java
```diff
@@ -0,0 +1,81 @@
...
+package org.apache.jackrabbit.vault.packaging.integration;
+
+import java.io.IOException;
+
+import javax.jcr.RepositoryException;
+import org.apache.jackrabbit.vault.packaging.PackageException;
+import org.junit.Test;
+
+
+public class MandatoryNode extends IntegrationTestBase {
```
kwin, line 27: Please rename to MandatoryNodeIT to make m-failsafe-p execute those ITs!

kwin (changes requested):

vault-core/src/test/java/org/apache/jackrabbit/vault/packaging/integration/MandatoryNode.java
reschke: done in https://github.com/apache/jackrabbit-filevault/pull/163/commits/750144a5b6fac1d51bd4eaa53b3ac76a16248152

vault-core/src/main/java/org/apache/jackrabbit/vault/fs/impl/io/DocViewSAXImporter.java
```diff
@@ -1295,6 +1298,33 @@ public void endElement(String uri, String localName, String qName) throws SAXExc
...
+            while (iter.hasNext()) {
+                Node sibling = iter.nextNode();
+                if (!sibling.isSame(child)) {
+                    Optional<NodeDefinition> childDef = ent.getApplicableChildNodeDefinition(sibling.getName(),
+                            sibling.getPrimaryNodeType());
+                    try {
+                        if (typeName.equals(childDef.get().getName())) {
+                            return true;
+                        }
+                    } catch (NoSuchElementException ignored) {
```
kwin, line 1319: I would rather do a check via `Optional.isPresent()` instead of relying on exceptions here.

vault-core/src/main/java/org/apache/jackrabbit/vault/fs/impl/io/DocViewSAXImporter.java
```diff
@@ -1295,6 +1298,33 @@ public void endElement(String uri, String localName, String qName) throws SAXExc
...
+                    }
+                }
+            }
+        } catch (NoSuchElementException ignored) {
```
kwin, line 1323: Same as above and do some logging in case the child node definition cannot be found

vault-core/src/main/java/org/apache/jackrabbit/vault/fs/impl/io/DocViewSAXImporter.java
```diff
@@ -1295,6 +1298,33 @@ public void endElement(String uri, String localName, String qName) throws SAXExc
         }
     }
 
+    private boolean hasSiblingWithSameType(Node child) throws RepositoryException {
+
+        Node parent = child.getParent();
+
+        try {
+            EffectiveNodeType ent = EffectiveNodeType.ofNode(parent);
+            String typeName = ent.getApplicableChildNodeDefinition(child.getName(), child.getPrimaryNodeType()).get().getName();
```
kwin, line 1307: You should consider the mixins as well!

vault-core/src/main/java/org/apache/jackrabbit/vault/fs/impl/io/DocViewSAXImporter.java
```diff
@@ -1295,6 +1298,33 @@ public void endElement(String uri, String localName, String qName) throws SAXExc
...
+
+            NodeIterator iter = parent.getNodes();
+            while (iter.hasNext()) {
+                Node sibling = iter.nextNode();
+                if (!sibling.isSame(child)) {
+                    Optional<NodeDefinition> childDef = ent.getApplicableChildNodeDefinition(sibling.getName(),
+                            sibling.getPrimaryNodeType());
```
kwin, line 1314: consider mixins

vault-core/src/main/java/org/apache/jackrabbit/vault/fs/impl/io/DocViewSAXImporter.java
```diff
@@ -1295,6 +1298,33 @@ public void endElement(String uri, String localName, String qName) throws SAXExc
...
+                    try {
+                        if (typeName.equals(childDef.get().getName())) {
```
kwin, line 1316: Shouldn't we rather compare childDef.get().getDeclaringNodeType().getName() (with a null check?)

kwin (changes requested):

vault-core/src/main/java/org/apache/jackrabbit/vault/fs/impl/io/DocViewSAXImporter.java
reschke: I wonder whether comparing the type is correct at all. Consider a mandatory child node of type "nt:resource", and a folder that contains both an "nt:file" and an "nt:folder". Their node types are different, but they both satisfy the parent's requirement.

vault-core/src/main/java/org/apache/jackrabbit/vault/fs/impl/io/DocViewSAXImporter.java
reschke: How are mixins relevant here?

vault-core/src/main/java/org/apache/jackrabbit/vault/fs/impl/io/DocViewSAXImporter.java
tripodsan: I the EffectiveNodeType already computes the, well, effective type, which is a merge of all the super and mixin types.

kwin: I am not talking about line 6 which takes into account mixins and inheritance, but about line 7 which retrieves the applicable mixin. IMHO this needs to parameterized with name, primary and mixin types otherwise the wrong applicable child node definition might be returned (the child node might actually be enforced to have a certain mixin instead of a primary)

vault-core/src/main/java/org/apache/jackrabbit/vault/fs/impl/io/DocViewSAXImporter.java
kwin: See above, if the parent enforces a child of a certain mixin this will only be returned by `getApplicableChildNodeDefinition` if the mixins are passed as well.

vault-core/src/main/java/org/apache/jackrabbit/vault/fs/impl/io/DocViewSAXImporter.java
kwin: Right, you need to take into account inheritance here.

vault-core/src/main/java/org/apache/jackrabbit/vault/fs/impl/io/DocViewSAXImporter.java
reschke: Again: how is this relevant here?

kwin: Example: 

1. Parent enforces mandatory child of mixin type m1
2. Parent contains two childs of that mixin 
3. `ent.getApplicableChildNodeDefinition(child.getName(), child.getPrimaryNodeType()`) will return `Optional.empty()` in that case
3. That leads to `hasSiblingWithSameType()` returning `false` which is obviously wrong here

tripodsan: but `ent.getApplicableChildNodeDefinition()` already operates on the merged node type....

vault-core/src/main/java/org/apache/jackrabbit/vault/fs/impl/io/DocViewSAXImporter.java
```diff
@@ -1295,6 +1297,37 @@ public void endElement(String uri, String localName, String qName) throws SAXExc
...
+    private boolean hasSiblingWithSameType(Node child) throws RepositoryException {
+
+        Node parent = child.getParent();
+        EffectiveNodeType ent = EffectiveNodeType.ofNode(parent);
+        Optional<NodeDefinition> childDef = ent.getApplicableChildNodeDefinition(child.getName(), child.getPrimaryNodeType());
+
+        if (!childDef.isPresent()) {
+            log.debug("no NodeDefinition for {}", child.getPath());
+        } else {
+            String typeName = childDef.get().getDeclaringNodeType().getName();
```
tripodsan, line 1309: why is the typename relevant? it's the mandatory type that needs to be compared.

vault-core/src/main/java/org/apache/jackrabbit/vault/fs/impl/io/DocViewSAXImporter.java
kwin: Seems that actually 1. is not supported by JCR: https://www.adobe.io/experience-manager/reference-materials/spec/jcr/2.0/3_Repository_Model.html#3.7.4.1%20Required%20Primary%20Node%20Types: Only primary types can be enforced but not mixins.

kwin:
Alternative approach in https://github.com/apache/jackrabbit-filevault/pull/181/files. @reschke Please have a look.

kwin:
Superseded by #181

PR #164, opened 2021-09-02
JCRVLT-557 - Testcase for creation of nodes calculating different nod…

DominikSuess:
…etype causing import failure

### DIFF
--- a/vault-core/src/test/resources/test-packages/test_implicit_definition.zip/META-INF/vault/filter.xml
+++ b/vault-core/src/test/resources/test-packages/test_implicit_definition.zip/META-INF/vault/filter.xml
@@ -1,0 +1,4 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<workspaceFilter version="1.0">
+    <filter root="/testroot/myfolder"/>
+</workspaceFilter>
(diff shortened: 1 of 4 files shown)


vault-core/src/test/resources/test-packages/test_implicit_definition.zip/META-INF/vault/filter.xml
```diff
@@ -0,0 +1,4 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<workspaceFilter version="1.0">
+    <filter root="/testroot/myfolder"/>
```
kwin, line 3: as `/testroot/myfolder` is contained in the package, the existing node in the repo is overwritten by an install. Since the package contents don't define a specific node type the default node type "nt:folder" is used. This is expected and is exactly the behaviour which was fixed by https://issues.apache.org/jira/browse/JCRVLT-142/https://issues.apache.org/jira/browse/JCRVLT-415. Why should the existing `nt:unstructured` node type being preserved here? IMHO the package is just incorrect!

kwin (changes requested):

vault-core/src/test/resources/test-packages/test_implicit_definition.zip/META-INF/vault/filter.xml
kwin: Here is the quote from @tripodsan

> a directory below a filter root w/o a .content.xml should have the nodetype nt:folder.

(https://issues.apache.org/jira/browse/JCRVLT-415?focusedCommentId=17046220&page=com.atlassian.jira.plugin.system.issuetabpanels%3Acomment-tabpanel#comment-17046220)

kwin:
Superseded by #179

PR #165, opened 2021-09-02
Testcase: JCRVLT-556 jackrabbit-packagetype validator wrongly complains about OSGI configurations in tools/config path

stefanseifert:
test case to reproduce problem


kwin:
@stefanseifert I agree that a proper IT is more useful. Have a look at https://github.com/apache/jackrabbit-filevault/pull/166.

stefanseifert:
obsolete with https://github.com/apache/jackrabbit-filevault/pull/166

PR #166, opened 2021-09-03
JCRVLT-556 evaluate primary type for non-file type nodes

kwin:
add integration test


kwin:
@stefanseifert Does this PR fix your example projects?

stefanseifert:
yes, indeed - thanks a lot!

PR #175, opened 2021-11-02
JCRVLT-557 - make default for overwriting node types configurable (OSGi)

reschke:


vault-core/src/main/java/org/apache/jackrabbit/vault/fs/impl/io/FolderArtifactHandler.java
```diff
@@ -73,6 +78,18 @@ public void setNodeType(String nodeType) {
         this.nodeType = nodeType;
     }
 
+    /**
+     * Sets whether primary type of folders should be overwritten
+     * 
+     * @param primary
```
kwin, line 84: wrong param and confusing description

kwin (approved):
LGTM except for the minor issue with the javadoc mentioned above and the 4 errors during javadoc creation (https://ci-builds.apache.org/blue/organizations/jenkins/Jackrabbit%2Ffilevault/detail/PR-175/1/pipeline). Please squash before merging.

reschke:
Thanks, will do.

reschke:
@kwin - two more questions:

1. Is it ok that I *removed* (not deprecated) " public FSPackageRegistry(@NotNull File homeDir, InstallationScope scope, @Nullable AbstractPackageRegistry.SecurityConfig securityConfig, boolean isStrict)"
2. Is there work left to do in "PackageTaskOptionsSerializer"?

kwin:
> 1. Is it ok that I _removed_ (not deprecated) " public FSPackageRegistry(@NotNull File homeDir, InstallationScope scope, @nullable AbstractPackageRegistry.SecurityConfig securityConfig, boolean isStrict)"

Yes, that is an impl class and should not be used anywhere outside FileVault (although this is only enforced with OSGi)

> 2. Is there work left to do in "PackageTaskOptionsSerializer"?
Good point, although only used in https://github.com/apache/sling-org-apache-sling-feature-extension-content/blob/7f982174eb877a3a0c7bd6b69862a3a11e9db137/src/main/java/org/apache/sling/feature/extension/content/ContentHandler.java#L82 AFAIK it would be good to serialize all options. We should have some more test coverage for this class as well.

reschke:
@kwin - PTAL at https://github.com/apache/jackrabbit-filevault/pull/175/commits/df384148c13b2550ec6e26323144c49fedb19e1c -- one puzzling thing here is that "PackageTaskOptionsSerializer" requires a method in "ImportOptions" that is deprecated; we should either undeprecate it, or see what needs to be changes in the serializer.

kwin:
@reschke Thanks for the update. LGTM, feel free to squash and merge. I am gonna take care of the deprecated method usage in ImportOptions.

PR #181, opened 2021-11-11
Bugfix/jcrvlt 549

kwin:


vault-core/src/test/java/org/apache/jackrabbit/vault/packaging/integration/MandatoryNodeIT.java
```diff
@@ -0,0 +1,62 @@
...
+        // check if newly added node exist
+        assertNodeExists("/libs/msm/wcm/rolloutconfigs/default/activate");
+        assertNodeExists("/libs/msm/wcm/rolloutconfigs/default/contentCopy");
+
+        // check if the only node is replaced
+        assertNodeMissing("/libs/msm/wcm/rolloutconfigs/activate/targetActivate");
+        assertNodeExists("/libs/msm/wcm/rolloutconfigs/activate/targetNewActivate");
+
+        // check if the only node is still there (although not part of 2nd package)
+        assertNodeExists("/libs/msm/wcm/rolloutconfigs/deactivate/targetDeactivate");
```
kwin, line 59: If I modify the code in DocViewSAXImporter.endNode() to remove mandatory child nodes this one suprisingly does not throw a ViolationConstraintException in Oak. @reschke any idea why deleting a mandatory child node is accepted?

kwin: https://issues.apache.org/jira/browse/OAK-9621

tripodsan: maybe we can relax the check and only reject the removal of the mandatory childnode, if it's a single node (eg, for a mandatory nt:resource). I think the _at least 1_ behaviour is rare...

kwin: I think JCR 2.0 spec is clear in this regard:
Mandatory residual child nodes definitions are invalid (and therefore not enforced),  mandatory non-residual child node definitions are valid. 

So FileVault should IMHO remove mandatory child nodes (even the last one) in case the definition is residual!

reschke (approved):

PR #182, opened 2021-12-02
JCRVLT-572 sling:OsgiConfig nodes in container packages produce errors, instead of being allowed

henrykuijpers:


kwin:
@henrykuijpers LGTM, thanks a lot. I enriched with an additional commit in #183 which also contains all your commits.

kwin:
Superseded by #183

PR #184, opened 2021-12-08
JCRVLT-574 pass connection options to VltContext

kwin:


reschke (approved):
looks good (but I did not test)

PR #186, opened 2021-12-13
JCRVLT-575: do not fail if home path for package registry contains a symlink

reschke:


kwin:
LGTM, can you add a small unit test showing your edge case?

vault-core/src/main/java/org/apache/jackrabbit/vault/packaging/registry/impl/FSInstallStateCache.java
```diff
@@ -55,8 +60,13 @@
     private final Path homeDir;
 
     public FSInstallStateCache(Path homeDir) throws IOException {
-        this.homeDir = homeDir;
-        Files.createDirectories(homeDir.toRealPath());
+        try {
+            this.homeDir = homeDir;
+            Files.createDirectories(homeDir.toRealPath());
+        } catch (FileAlreadyExistsException ex) {
+            log.error("on createDirectories for " + homeDir + " (real path: " + homeDir.toRealPath() + ")", ex);
```
kwin, line 67: Let us rather wrap the original exception and add this detail message. Both throwing and logging is an antipattern and leads to redundant log entries.

reschke: Any preference about that Exception to wrap into?

kwin: A generic IOException would be fine for me.

reschke:
> LGTM, can you add a small unit test showing your edge case?

Not easily, because I can't create symlinks on Windows.

kwin:
It is fine to have certain unit tests just for some OS (we always check against Mac, Linux and Windows). Just use https://junit.org/junit4/javadoc/4.13/org/junit/Assume.html to ignore a test on Windows.

PR #187, opened 2021-12-15
JCRVLT-576 make vault-validation runnable in OSGi container

kwin:
include all non-exported classes/packages from Vault Core
embed jcr2spi


kwin:
@ankitaagar Please check if this approach works for you.

PR #188, opened 2021-12-15
JCRVLT-575: handle symlinks in homeDir path

reschke:
Functional changes besides logging:

1. do not call "createDirectories" when the path already exists (this avoids a failure when the last path segment is a symlink)
2. when walking the homeDir, follow symlinks

### DIFF
--- a/vault-core/src/main/java/org/apache/jackrabbit/vault/packaging/registry/impl/FSInstallStateCache.java
+++ b/vault-core/src/main/java/org/apache/jackrabbit/vault/packaging/registry/impl/FSInstallStateCache.java
@@ -18,6 +18,7 @@ package org.apache.jackrabbit.vault.packaging.registry.impl;
 
 import java.io.IOException;
 import java.io.UncheckedIOException;
+import java.nio.file.FileVisitOption;
 import java.nio.file.Files;
 import java.nio.file.Path;
 import java.util.AbstractMap;
@@ -30,6 +31,8 @@ import java.util.stream.Stream;
 
 import org.apache.jackrabbit.vault.packaging.PackageId;
 import org.jetbrains.annotations.NotNull;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
 
 /**
  * Persisted cache of all {@link FSInstallState} objects for all packages in a registry.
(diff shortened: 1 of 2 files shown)


DominikSuess (approved):
lgtm - validated changes under the conditions that provoked failures before (registry from a symlinked locatioN)

kwin:
@reschke Can you add a test leveraging https://docs.oracle.com/javase/8/docs/api/java/nio/file/Files.html#createSymbolicLink-java.nio.file.Path-java.nio.file.Path-java.nio.file.attribute.FileAttribute...- to make sure it doesn't regress in the future?

vault-core/src/main/java/org/apache/jackrabbit/vault/packaging/registry/impl/FSInstallStateCache.java
```diff
@@ -53,10 +58,14 @@
     private Map<Path, PackageId> pathIdMapping = new ConcurrentHashMap<>();
 
     private final Path homeDir;
-    
+
     public FSInstallStateCache(Path homeDir) throws IOException {
         this.homeDir = homeDir;
-        Files.createDirectories(homeDir);
+        log.debug("checking for presence of {} - exists {} - isDirectory {}", homeDir, Files.exists(homeDir), Files.isDirectory(homeDir));
+        if (Files.notExists(homeDir)) {
```
kwin, line 65: What happens if the path `homeDir` is a file and not a directory?

cziegeler (approved):

vault-core/src/main/java/org/apache/jackrabbit/vault/packaging/registry/impl/FSInstallStateCache.java
reschke: Ok, in that case we should fail. Will do.

reschke:
> @reschke Can you add a test leveraging https://docs.oracle.com/javase/8/docs/api/java/nio/file/Files.html#createSymbolicLink-java.nio.file.Path-java.nio.file.Path-java.nio.file.attribute.FileAttribute...- to make sure it doesn't regress in the future?

Is there an existing test class that I can extend? (Note that this is a regression from 3.4.0 which works with symlinks just fine, and we really really want to finally upgrade)

kwin:
> Is there an existing test class that I can extend?

There is org.apache.jackrabbit.vault.packaging.registry.impl.FSInstallStateTest (pretty obvious, right?)

> and we really really want to finally upgrade

Yes, I am not saying that this blocks the merge, but it has regressed before because there were no test covering that. Let's do better this time!

reschke:
@kwin - please see https://github.com/apache/jackrabbit-filevault/pull/188/commits/b88b2838cf8a53ba83d20c51e821ca3931f5476b

kwin (approved):
LGTM, thanks

PR #202, opened 2022-01-17
JCRVLT-596: add OSGi config for default IdConflictPolicy

reschke:


vault-core/src/main/java/org/apache/jackrabbit/vault/packaging/impl/PackagingImpl.java
```diff
@@ -106,6 +106,9 @@ public PackagingImpl() {
 
         @AttributeDefinition(description = "Whether to overwrite the primary type of folders")
         boolean overwritePrimaryTypesOfFolders() default true;
+
+        @AttributeDefinition(description = "Default IdConflictPolicy")
```
kwin, line 110: Please use the enum values instead (not free-text String), that also prevents type conversion.

reschke: For some reason I though this would be hard, but apparently it's not :-)

vault-core/src/main/java/org/apache/jackrabbit/vault/packaging/impl/JcrPackageImpl.java
```diff
@@ -386,6 +386,7 @@ private void extract(Set<PackageId> processed, ImportOptions options, boolean cr
             // MAX_VALUE disables saving completely, therefore we have to use a lower value!
             opts.setAutoSaveThreshold(Integer.MAX_VALUE - 1);
         }
+        opts.setIdConflictPolicy(mgr.getDefaultIdConflictPolicy());
```
kwin, line 389: This must be conditional, only override if not yet set!

kwin (changes requested):

vault-core/src/main/java/org/apache/jackrabbit/vault/packaging/impl/JcrPackageImpl.java
reschke: so that needs extra code in ImportOptions, right?

reschke: I tried to embed the logic in ImportOptions, and extended "toString()" and friends accordingly. Now ExecutionPlanBuilderImplTest.testSaveAndLoad fails - what do I need to adjust?

reschke: Changed to use null for not set. Please review.

reschke:
@kwin  - thanks for the help! Should I commit?

vault-core/src/main/java/org/apache/jackrabbit/vault/packaging/impl/JcrPackageImpl.java
kwin: Line 392 must be removed!

kwin (changes requested):

kwin (approved):

PR #203, opened 2022-01-31
JCRVLT-601 improve documentation on namespace handling

kwin:


tripodsan (approved):

reschke (approved):
Looks ok from a quick glance, but I'm not that familiar with this area.

PR #204, opened 2022-02-08
JCRVLT-604 - RepositoryCopier doesn't allow Property Based Filtering

clotton:
**JIRA:**
https://issues.apache.org/jira/browse/JCRVLT-604

**DESCRIPTION:**
- Utilize the WorkspaceFilter's includesProperty util method to determine if a given property should be ignored during copy

### DIFF
--- a/vault-core/src/main/java/org/apache/jackrabbit/vault/util/RepositoryCopier.java
+++ b/vault-core/src/main/java/org/apache/jackrabbit/vault/util/RepositoryCopier.java
@@ -266,7 +266,7 @@ public class RepositoryCopier {
         currentSize = 0;
         totalSize = 0;
         start = System.currentTimeMillis();
-        
+
         AutoSave autoSave = new AutoSave();
         autoSave.setThreshold(getBatchSize());
         autoSave.setTracker(new ProgressTracker(tracker));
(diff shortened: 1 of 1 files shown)


clotton:
@kwin Here's a PR as requested.  Note I don't have the permissions to assign people to review it.

vault-core/src/main/java/org/apache/jackrabbit/vault/util/RepositoryCopier.java
```diff
@@ -410,11 +410,17 @@ private void copy(AutoSave autoSave, Node src, Node dstParent, String dstName, b
                             currentSize+=s;
                         }
                     } else {
-                        Value v = p.getValue();
-                        dst.setProperty(pName, v);
-                        long s= p.getLength();
-                        totalSize+=s;
-                        currentSize+=s;
+                        // Filter out unwanted properties.  Remove them from the destination if they are excluded by filter
```
kwin, line 413: The same logic would need to be applied to `sysCopy` which uses JCR System View to copy nodes from src to dst. This requires some more code changes as you probably need to implement a filtering ContentHandler (based on http://www.saxproject.org/apidoc/org/xml/sax/helpers/XMLFilterImpl.html).

kwin (changes requested):
Looks good but is not complete yet (see inline comments). Also it would be good to add a test checking the behaviour.

vault-core/src/main/java/org/apache/jackrabbit/vault/util/RepositoryCopier.java
clotton: Thanks @kwin does the current sysCopy method already filter out nodes somehow, but it is missing property based filtering?  I'm a little lost looking at the current [sysCopy](https://github.com/apache/jackrabbit-filevault/blob/master/vault-core/src/main/java/org/apache/jackrabbit/vault/util/RepositoryCopier.java#L505) having never been in here before.

clotton: @kwin it's not clear to me how we would intercept the ContentHandler found in syscopy, which is an ImportHandler defined in oak-run.   A new implementation would, perhaps, need to extend ImportHandler to apply filtering, and the pass it on to the exportSystemView call as done already.

If you have any guidance, or examples, that would be greatly appreciated.  If you'd like to schedule some time to discuss this, I'd welcome it.

PR #206, opened 2022-02-14
JCRVLT-605 don't overwrite filtered protected properties during import

kwin:


vault-core/src/main/java/org/apache/jackrabbit/vault/fs/impl/io/DocViewImporter.java
```diff
@@ -110,23 +111,30 @@
...
         Set<Name> props = new HashSet<>();
         props.add(NameConstants.JCR_PRIMARYTYPE);
         props.add(NameConstants.JCR_MIXINTYPES);
         props.add(NameConstants.JCR_UUID);
+        PROTECTED_PROPERTIES_CONSIDERED_FOR_UPDATED_NODES = Collections.unmodifiableSet(props);
         props.add(NameConstants.JCR_ISCHECKEDOUT);
         props.add(NameConstants.JCR_BASEVERSION);
         props.add(NameConstants.JCR_PREDECESSORS);
         props.add(NameConstants.JCR_SUCCESSORS);
         props.add(NameConstants.JCR_VERSIONHISTORY);
```
kwin, line 133: @tripodsan Any idea why these properties are imported via sysview for new nodes but
a) never exported through DocViewSaxFormatter
b) never imported for an update

PR #214, opened 2022-03-07
fix Reproducible Builds issues

hboutemy:
see current status of last releases: https://github.com/jvm-repo-rebuild/reproducible-central#org.apache.jackrabbit.vault:jackrabbit-filevault

issues are only on examples, using old plugins => this PR upgrades


kwin:
@hboutemy Thanks a lot.

PR #225, opened 2022-06-09
JCRVLT-598 introduce new IdConflictPolicy.LEGACY which restores the

kwin:


reschke:
Looks good so far. Will try it on my test content later today.

reschke (approved):
I deployed this on my test system, and can confirm that LEGACY behaves as expected.

I did not do a code review; I'll trust you with that :-).

PR #227, opened 2022-06-13
JCRVLT-637: fix support for non-docview files with additional properties

Buuhuu:
The DocumentViewParserValidator returns a binary file back to the ValidationExecutor as node path with a line number of 0.
The ValidationExecutor considers this state as docview xml case.

With this change the ValidationExecutor checks the returned nodePaths and lineNubmers and if only one nodePath (the file) is returned with lineNumber 0, it handles it the same way as if nothing would have been returned (existing logic).

### DIFF
--- a/vault-validation/src/main/java/org/apache/jackrabbit/vault/validation/ValidationExecutor.java
+++ b/vault-validation/src/main/java/org/apache/jackrabbit/vault/validation/ValidationExecutor.java
@@ -309,10 +309,13 @@ public final class ValidationExecutor {
                                 enrichedMessages.addAll(ValidationViolation.wrapMessages(entry.getKey(), messages, filePath, basePath, null, 0, 0));
                             }
                         } 
+
                         // only do it if we haven't collected node paths from a previous run
-                        if (nodePathsAndLineNumbers.isEmpty()) {
-                            // convert file name to node path
-                            String nodePath = filePathToNodePath(filePath);
+                        String nodePath = filePathToNodePath(filePath);
+                        boolean treatedAsBinaryFile = nodePathsAndLineNumbers.size() == 1 &&
+                            nodePathsAndLineNumbers.getOrDefault(nodePath, Integer.MIN_VALUE) == 0;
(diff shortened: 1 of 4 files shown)


kwin:
@Buuhuu Thanks for the report and the PR. While I am still trying to understand it would it be possible to enhance this PR by an additional IT (with the example from the JIRA description), the unit test is very artificial.

Buuhuu:
Sure. 

Let me explain the issue in more detail. In [validateGenericJcrData](https://github.com/apache/jackrabbit-filevault/blob/master/vault-validation/src/main/java/org/apache/jackrabbit/vault/validation/ValidationExecutor.java#L313-L319) the `ValidationExecutor` considers a `filePath` as non-docview file (`isDocViewXml = false`) only when the `GenericJcrDataValidator` implementation returns with an empty `nodePathsAndLineNumbers` list. The [DocumentViewParserValidator](https://github.com/apache/jackrabbit-filevault/blob/jackrabbit-filevault-3.6.0/vault-validation/src/main/java/org/apache/jackrabbit/vault/validation/spi/impl/DocumentViewParserValidator.java#L105) however, does not return an empty list in this case but adds the filePath to the list itself (duplicated logic). This causes the [NodeTypeValidator](https://github.com/apache/jackrabbit-filevault/blob/jackrabbit-filevault-3.6.0/vault-validation/src/main/java/org/apache/jackrabbit/vault/validation/spi/impl/nodetype/NodeTypeValidator.java#L342) to skip the file in this stage, and so the `jcr:data` attribute is missing on the `nt:resource` node. 

We could also remove this [line](https://github.com/apache/jackrabbit-filevault/blob/jackrabbit-filevault-3.6.0/vault-validation/src/main/java/org/apache/jackrabbit/vault/validation/spi/impl/DocumentViewParserValidator.java#L105), but there is a unit test that validates this behaviour of the DocumentViewParserValidator so I assume that would be a breaking change.

kwin:
Superseded by #228

PR #228, opened 2022-06-15
JCRVLT-637

kwin:


Buuhuu:
Yes, as [said earlier](https://github.com/apache/jackrabbit-filevault/pull/227#issuecomment-1156124727), this should work

> We could also remove this [line](https://github.com/apache/jackrabbit-filevault/blob/jackrabbit-filevault-3.6.0/vault-validation/src/main/java/org/apache/jackrabbit/vault/validation/spi/impl/DocumentViewParserValidator.java#L105), but there is a unit test that validates this behaviour of the DocumentViewParserValidator so I assume that would be a breaking change.

With that you can remove `testGenericJcrDataWithBinaryFileDetected()` as this should now be covered by `testGenericJcrData()` already.

kwin:
> With that you can remove testGenericJcrDataWithBinaryFileDetected() as this should now be covered by testGenericJcrData() already.
Right, gonna remove the former.

> I don't understand how these changes are related
https://github.com/apache/jackrabbit-filevault/pull/228/files#diff-d467e47245ea852f2e8ae6cc036e413bc2098a0e5be790a56eb238e3a17f3b41R375-R383

The problem was that `testGenericJcrDataWithBinaryFileDetected()` was never failing for me when called from the IDE, as in that case the validators `genericjcrdataid2` and `jackrabbit-docviewparser` had the wrong order (the latter must always come first). As the new map originally didn't retain any order, this fix was necessary to make the test and probably also the validation at runtime always behave the same (irrespective of underlying Map implementations).

PR #240, opened 2022-07-28
JCRVLT-646 Potential javax.jcr.nodetype.ConstraintViolationException while deserializing extended file aggregates

kwin:


reschke (commented):
I believe we should strip the contents further down so that no content with non-Apache copyrights is left...

kwin:
> I believe we should strip the contents further down

Done now. Please have another look.

reschke (commented):
Looks good, will run IT in AEM tomorrow.

PR #241, opened 2022-08-03
JCRVLT-647 restore old save() interval after failed save() attempts

kwin:
Verify that counters behave correctly


reschke (approved):

reschke (changes requested):
I'm not sure why but it seems that in my test, it now saves after 1024 nodes, then 2048 nodes, and then is back to a threshold of 1.

kwin:
> ... back to a threshold of 1.

This should be fixed now in https://github.com/apache/jackrabbit-filevault/pull/241/commits/4d4407c8b774a0be9b2119937acc2a318eb5e50e

Still the save() won't recover in that case, i.e. the package install won't succeed.

reschke (approved):

PR #243, opened 2022-08-04
JCRVLT-647: Restore old intermediate save() interval after failed save

reschke:


reschke:
...1024 instead of 10.

kwin:
@reschke See #241

kwin:
Superseded by #241

PR #245, opened 2022-09-24
JCRVLT-658: Importer fails for invalid namespace names

reschke:


vault-core/src/test/resources/test-packages/badnamespacenames.zip/META-INF/vault/config.xml
```diff
@@ -0,0 +1,93 @@
+<!--
```
kwin, line 1: file is default and not necessary

vault-core/src/test/resources/test-packages/badnamespacenames.zip/META-INF/vault/definition/.content.xml
```diff
@@ -0,0 +1,25 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
kwin, line 1: file is not necessary

vault-core/src/test/resources/test-packages/badnamespacenames.zip/META-INF/vault/nodetypes.cnd
```diff
@@ -0,0 +1,17 @@
+<'vlt'='http://www.day.com/jcr/vault/1.0'>
```
kwin, line 1: file is not necessary

vault-core/src/main/java/org/apache/jackrabbit/vault/fs/impl/io/DocViewImporter.java
```diff
@@ -1200,28 +1200,37 @@ private boolean isPropertyProtected(@NotNull EffectiveNodeType effectiveNodeType
         return effectiveNodeType.getApplicablePropertyDefinition(npResolver.getJCRName(docViewProperty.getName()), docViewProperty.isMultiValue(), docViewProperty.getType()).map(PropertyDefinition::isProtected).orElse(false);
     }
 
+    private String nameToPathComponent(Name name) throws RepositoryException {
```
kwin, line 1203: maybe `getLexicalForm` as method name (https://developer.adobe.com/experience-manager/reference-materials/spec/jcr/2.0/3_Repository_Model.html#3.2.5%20Lexical%20Form%20of%20JCR%20Names) which would allow both expanded and qualified form.

kwin (changes requested):

vault-core/src/main/java/org/apache/jackrabbit/vault/fs/impl/io/DocViewImporter.java
reschke: sounds right to me

kwin:
I guess we also need adaptations in https://github.com/apache/jackrabbit-filevault/blob/415a858cc5816f58b8010172360a69703e58b953/vault-core/src/main/java/org/apache/jackrabbit/vault/util/DocViewProperty2.java#L519.

kwin:
The IT fails because the bad namespaces are not registered. Either add those via CND file or programmatically in the IT.

vault-core/src/main/java/org/apache/jackrabbit/vault/fs/impl/io/DocViewImporter.java
```diff
@@ -1200,28 +1200,37 @@ private boolean isPropertyProtected(@NotNull EffectiveNodeType effectiveNodeType
         return effectiveNodeType.getApplicablePropertyDefinition(npResolver.getJCRName(docViewProperty.getName()), docViewProperty.isMultiValue(), docViewProperty.getType()).map(PropertyDefinition::isProtected).orElse(false);
     }
 
+    private String getLexicalForm(Name name) throws RepositoryException {
```
reschke, line 1203: OTOH, we could just always use "getJCRName()"...

reschke:
> The IT fails because the bad namespaces are not registered. Either add those via CND file or programmatically in the IT.

They only fail for the property; no. The import *does* register the namespaces...

vault-core/src/main/java/org/apache/jackrabbit/vault/fs/impl/io/DocViewImporter.java
kwin: IMHO this additional lookup costs some performance.

kwin (approved):
LGTM, thanks!

PR #247, opened 2022-10-06
JCRVLT-661 Document how authorizables and acls are exported/imported

kwin:


kwin:
@tripodsan Can you have a look?

PR #250, opened 2022-10-10
JCRVLT-662: JcrPackageRegistry.remove() leaks temp file

mreutegg:
Fix and test to validate

### DIFF
--- a/vault-core/src/main/java/org/apache/jackrabbit/vault/packaging/registry/impl/JcrPackageRegistry.java
+++ b/vault-core/src/main/java/org/apache/jackrabbit/vault/packaging/registry/impl/JcrPackageRegistry.java
@@ -651,22 +651,21 @@ public class JcrPackageRegistry extends AbstractPackageRegistry {
     @SuppressWarnings("resource")
     @Override
     public void remove(@NotNull PackageId id) throws IOException, NoSuchPackageException {
-        JcrRegisteredPackage pkg = (JcrRegisteredPackage) open(id);
-        if (pkg == null) {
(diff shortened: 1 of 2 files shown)


vault-core/src/main/java/org/apache/jackrabbit/vault/packaging/registry/impl/JcrPackageRegistry.java
```diff
@@ -655,8 +655,8 @@ public void remove(@NotNull PackageId id) throws IOException, NoSuchPackageExcep
         if (pkg == null) {
             throw new NoSuchPackageException().setId(id);
         }
-        JcrPackage pack = pkg.getJcrPackage();
         try {
+            JcrPackage pack = pkg.getJcrPackage();
```
kwin, line 659: you can use [try with resources](https://docs.oracle.com/javase/tutorial/essential/exceptions/tryResourceClose.html) as `JcrPackage` implements `AutoClosable`.

mreutegg: Ah, missed that one. Though, it seems it is not necessary in this case. Closing `JcrRegisteredPackage` already closes the `JcrPackage` it holds.

mreutegg: Done. Also put the `snap` `JcrPackage` into a try with resources.

kwin (approved):
LGTM

PR #258, opened 2022-12-13
JCRVLT-670: Upgrade to commons-collections4

reschke:


parent/pom.xml
```diff
@@ -502,7 +502,7 @@ Bundle-Category: jackrabbit
             <dependency>
                 <groupId>org.apache.commons</groupId>
                 <artifactId>commons-collections4</artifactId>
-                <version>4.1</version>
```
kwin, line 505: Why is that upgrade necessary? That makes it incompatible with older distributions like AEM 6.4.

vault-core/pom.xml
```diff
@@ -182,6 +182,11 @@
             <artifactId>org.osgi.framework</artifactId>
             <scope>provided</scope>
         </dependency>
+        <dependency>
```
kwin, line 185: Having this dependency (although not new) for only one usage seems not really good.
Let us rather try to get rid of commons collection4 dependency in vault core.

vault-validation/pom.xml
```diff
@@ -124,6 +124,12 @@
             <version>${project.version}</version>
             <scope>compile</scope>
         </dependency>
+        <dependency>
+            <groupId>org.apache.commons</groupId>
+            <artifactId>commons-collections4</artifactId>
```
kwin, line 129: I don't see it being used in this module....

parent/pom.xml
reschke: FV is already incompatible with AEM 6.4 (unless I'm missing something).

vault-core/pom.xml
reschke: Good point, and I consirered that myself. It seems only be used for avoiding string duplicates in memory. Maybe that's something not relevant anymore in today's VMs.

vault-validation/pom.xml
reschke: Not having it here causes a test failure.

parent/pom.xml
kwin: Really? At least from a dependency pov this is checked during the build. Why do you think otherwise?

reschke: Because there are incompatibilties in AEM code that require code changes to be compatible with Filevault (newer than 3.4.0). You can probably deploy it, but there may be hard-to-identify problems later on.

reschke: FWIW, see https://helpx.adobe.com/support/programs/eol-matrix.html - AEM 6.4 will be out of extended support in five months from now.

kwin: It is not necessarily about FileVault Core but other bundles like RCP.

reschke: Hmmm. RCP doesn't use commons-collections4 anyway, right?

In any case I agree that getting rid of the dependency would be preferable.

PR #260, opened 2022-12-15
JCRVLT-672 use mavens ComparableVersion for package Version compareTo

stoerr:
Compare https://issues.apache.org/jira/browse/JCRVLT-672 : the advantage is that this properly orders e.g. -SNAPSHOT versions of packages before: previously, 1.2.3 < 1.2.3-SNAPSHOT , but normally you'd create 1.2.3-SNAPSHOT and only later the actual release 1.2.3 . When Version.compareTo is used to check whether a package is "newer" (as e.g. Adobe AEM does), it could refuse installing release 1.2.3 after 1.2.3-SNAPSHOT is installed, which is wrong.

This changes the logic of Version.compareTo to use the maven logic  https://maven.apache.org/pom.html#Version_Order_Specification  .

### DIFF
--- a/vault-core/src/main/java/org/apache/jackrabbit/vault/packaging/Version.java
+++ b/vault-core/src/main/java/org/apache/jackrabbit/vault/packaging/Version.java
@@ -166,7 +166,7 @@ public class Version implements Comparable<Version> {
      * @see "https://cwiki.apache.org/confluence/display/MAVENOLD/Versioning"
      * @see "https://semver.org/spec/v1.0.0.html"
      */
-    public int compareTo(@NotNull Version o) {
+    public int compareTo(Version o) {
         ComparableVersion thisVersion = new ComparableVersion(toString());
         ComparableVersion otherVersion = new ComparableVersion(o.toString());
         return thisVersion.compareTo(otherVersion);
--- a/vault-core/src/main/java/org/apache/jackrabbit/vault/packaging/Version.java
+++ b/vault-core/src/main/java/org/apache/jackrabbit/vault/packaging/Version.java
@@ -140,9 +140,8 @@ public class Version implements Comparable<Version> {
      * <a href="https://maven.apache.org/">Apache Maven</a>. A version usually consists  out of three numerical parts separated by dot -
      * major version, minor version and patch level, and can be followed by a dash and a qualifier like SNAPSHOT.
      * Version numbers can also consist of fewer or more parts (numerical, string, ....).
-     * If the comparison is not resolved by comparing the numbers, the algorith resorts to the qualifier - see
-     * <a href="https://cwiki.apache.org/confluence/display/MAVENOLD/Versioning">"Versioning" on Maven Wiki</a>
-     * for details.
+     * The comparison of the versions conforms to the
+     * <a href="https://maven.apache.org/pom.html#version-order-specification">Maven version order specification</a>.
      *
      * <pre>
      * Some examples:
@@ -163,8 +162,8 @@ public class Version implements Comparable<Version> {
      * @return  a negative integer, zero, or a positive integer as this version
      *		is less than, equal to, or greater than the specified version.
      *
-     * @see "https://cwiki.apache.org/confluence/display/MAVENOLD/Versioning"
-     * @see "https://semver.org/spec/v1.0.0.html"
+     * @see <a href="https://maven.apache.org/pom.html#version-order-specification">Maven version order specification</a>
+     * @see <a href="https://semver.org/spec/v1.0.0.html">Semantic Versioning 1.0.0</a>
      */
     public int compareTo(Version o) {
         ComparableVersion thisVersion = new ComparableVersion(toString());
--- a/vault-core/src/main/java/org/apache/jackrabbit/vault/packaging/Version.java
+++ b/vault-core/src/main/java/org/apache/jackrabbit/vault/packaging/Version.java
@@ -20,6 +20,7 @@ package org.apache.jackrabbit.vault.packaging;
 import java.util.ArrayList;
 
 import org.apache.jackrabbit.util.Text;
+import org.apache.maven.artifact.versioning.ComparableVersion;
 import org.jetbrains.annotations.NotNull;
 import org.jetbrains.annotations.Nullable;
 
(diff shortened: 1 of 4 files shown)


vault-core/src/main/java/org/apache/jackrabbit/vault/packaging/Version.java
```diff
@@ -135,44 +136,40 @@ public String[] getNormalizedSegments() {
...
-                return v1 - v2;
-            } catch (NumberFormatException e) {
-                // no numbers, use string compare
-                return strCompare;
-            }
-        }
-        return segments.length - oSegs.length;
+    public int compareTo(@NotNull Version o) {
+        ComparableVersion thisVersion = new ComparableVersion(toString());
+        ComparableVersion otherVersion = new ComparableVersion(o.toString());
```
stoerr, line 171: Here is a tradeoff: if version comparisons are infrequent, its better to do it this way. If version comparisons are frequent, it'd be better to create the ComparableVersion during object construction and keep it as a final field. What do you think, @kwin ?

kwin:
@stoerr Can you rebase to see if that fixes the Jenkins error?

vault-core/src/main/java/org/apache/jackrabbit/vault/packaging/Version.java
```diff
@@ -135,44 +136,40 @@ public String[] getNormalizedSegments() {
     }
 
     /**
-     * Compares this version to the given one, segment by segment without any special
-     * "SNAPSHOT" handling.
+     * Compares this version to the given one. The comparison is compatible to the ordering used by
+     * <a href="https://maven.apache.org/">Apache Maven</a>. It version consists normally from 3 numbers -
```
kwin, line 140:
```suggestion
     * <a href="https://maven.apache.org/">Apache Maven</a>. A version usually consists  out of three numerical parts separated by dot -
```

vault-core/src/main/java/org/apache/jackrabbit/vault/packaging/Version.java
```diff
@@ -135,44 +136,40 @@ public String[] getNormalizedSegments() {
...
+     * major version, minor version and patch level, and can be followed by a dash and a qualifier like SNAPSHOT.
+     * Version numbers can also consist of fewer or more numbers.
```
kwin, line 142:
```suggestion
     * Version numbers can also consist of fewer or more parts (numerical, string, ....).
```

vault-core/src/main/java/org/apache/jackrabbit/vault/packaging/Version.java
```diff
@@ -135,44 +136,40 @@ public String[] getNormalizedSegments() {
...
+     * If the comparison is not resolved by comparing the numbers, the algorith resorts to the qualifier - see
+     * <a href="https://cwiki.apache.org/confluence/display/MAVENOLD/Versioning">"Versioning" on Maven Wiki</a>
```
kwin, line 144: This resource is really outdated, rather link something more recent like https://maven.apache.org/pom.html#version-order-specification

vault-core/src/main/java/org/apache/jackrabbit/vault/packaging/Version.java
```diff
@@ -135,44 +136,40 @@ public String[] getNormalizedSegments() {
     }
 
     /**
-     * Compares this version to the given one, segment by segment without any special
-     * "SNAPSHOT" handling.
+     * Compares this version to the given one. The comparison is compatible to the ordering used by
+     * <a href="https://maven.apache.org/">Apache Maven</a>. It version consists normally from 3 numbers -
+     * major version, minor version and patch level, and can be followed by a dash and a qualifier like SNAPSHOT.
+     * Version numbers can also consist of fewer or more numbers.
+     * If the comparison is not resolved by comparing the numbers, the algorith resorts to the qualifier - see
```
kwin, line 143: This sentence feels fuzzy, I would just link out to the version order specification from Maven: https://maven.apache.org/pom.html#version-order-specification

vault-core/src/main/java/org/apache/jackrabbit/vault/packaging/Version.java
```diff
@@ -135,44 +136,40 @@ public String[] getNormalizedSegments() {
...
      *
-     * @param o the other version
+     * Please note that this comparison does not use the exact segmentation presented in {@link #getNormalizedSegments()},
+     * but applies the maven comparison algorithm to the string representation {@link #toString()}.
+     *
+     * @param o the other version, not null
      * @return  a negative integer, zero, or a positive integer as this version
      *		is less than, equal to, or greater than the specified version.
+     *
+     * @see "https://cwiki.apache.org/confluence/display/MAVENOLD/Versioning"
```
kwin, line 166: This is an invalid syntax: https://docs.oracle.com/en/java/javase/17/docs/specs/javadoc/doc-comment-spec.html#see

vault-core/src/main/java/org/apache/jackrabbit/vault/packaging/Version.java
```diff
@@ -135,44 +136,40 @@ public String[] getNormalizedSegments() {
...
+     * @see "https://semver.org/spec/v1.0.0.html"
```
kwin, line 167: This is an invalid syntax: https://docs.oracle.com/en/java/javase/17/docs/specs/javadoc/doc-comment-spec.html#see

vault-core/src/test/java/org/apache/jackrabbit/vault/packaging/VersionRangeTest.java
```diff
@@ -105,9 +105,18 @@ public void testRangeWithEmptyVersionParameter() {
         assertFalse("[1.0,2.0] excludes empty version", vr.isInRange(Version.EMPTY));
     }
 
+    /**
+     * A snapshot precedes the released version, so 1.0-SNAPSHOT not contained in the range [1.0,2.0)
```
kwin, line 109:
```suggestion
     * A snapshot precedes the released version, so 1.0-SNAPSHOT is not contained in the range [1.0,2.0)
```

vault-core/src/test/java/org/apache/jackrabbit/vault/packaging/VersionRangeTest.java
```diff
@@ -105,9 +105,18 @@ public void testRangeWithEmptyVersionParameter() {
...
+     * - but the bound can be a snapshot.
```
kwin, line 110:
```suggestion
     * - but the boundary can be a snapshot.
```

vault-core/src/test/java/org/apache/jackrabbit/vault/packaging/VersionRangeTest.java
```diff
@@ -105,9 +105,18 @@ public void testRangeWithEmptyVersionParameter() {
...
+     *
+     * @see "https://github.com/apache/maven/blob/maven-3.8.6/maven-artifact/src/test/java/org/apache/maven/artifact/versioning/VersionRangeTest.java#L657"
```
kwin, line 112: wrong syntax

kwin (changes requested):

vault-core/src/main/java/org/apache/jackrabbit/vault/packaging/Version.java
stoerr: Ah, thanks! I've been doing that wrong for ages. 8-}

stoerr:
@kwin Puzzling. The build worked after the rebase, but after the review changes it doesn't work, again. And it still shows "1 change requested" although all review comments have been addressed. 🤷

kwin:
Build failed due to invalid javadoc:

`[ERROR] F:\jenkins\jenkins-home\712657a4\workspace\Jackrabbit_filevault_PR-260\vault-core\src\main\java\org\apache\jackrabbit\vault\packaging\Version.java:144: error: unexpected end tag: </a>
[ERROR]      * <a href="https://maven.apache.org/pom.html#version-order-specification">Maven version order specification</a></a>.`

(https://ci-builds.apache.org/blue/rest/organizations/jenkins/pipelines/Jackrabbit/pipelines/filevault/branches/PR-260/runs/6/nodes/37/steps/75/log/?start=0).

Regarding 
> And it still shows "1 change requested" although all review comments have been addressed. 🤷

only the reviewer decides about whether everything has been addressed or not (which kind of makes sense). This is GitHub default for PRs.

vault-core/src/main/java/org/apache/jackrabbit/vault/packaging/Version.java
kwin: field is better as the size is neglectible but the parsing in the constructor is complex

stoerr:
Ouch! I'd never ever would have found that problem in the logs. 😲  Thank you!

kwin (approved):

kwin:
Merged a slightly modified version in https://github.com/apache/jackrabbit-filevault/commit/1442a06b288b0f90f3e78ccd5ad0291f977675bb. @stoerr Thanks for the efforts.

stoerr:
Thank you as well!

PR #269, opened 2023-01-16
JCRVLT-680: vault-validation - explicitly add commons-collections4 as test dependency

reschke:


vault-validation/pom.xml
```diff
@@ -247,6 +247,12 @@
             <artifactId>slf4j-simple</artifactId>
             <scope>test</scope>
         </dependency>
+        <dependency>
+            <!-- see JCRVLT-680 -->
+            <groupId>org.apache.commons</groupId>
+            <artifactId>commons-collections4</artifactId>
+            <scope>test</scope>
```
kwin, line 254: IMHO this is not only a test scope dependency, but really a runtime one (i.e. client using the validation API will need it). Therefore this should have scope "provided". All consumers probably need a fix as well as "provided" is not transitive.

kwin (changes requested):

vault-validation/pom.xml
reschke: The root cause is that vault-validation sets jackrabbit.version to 2.16.0, and that version predates the switch to commons-collections4.

kwin (approved):

PR #273, opened 2023-02-13
JCRVLT-684: test coverage and fix for bug in mixin restoration

reschke:


reschke:
Might need some more documentation, but here we go.

reschke:
@kwin - running the test in isolation works over here, but when running full tests it fails; is that a concurrency issue? (me confused)

reschke:
So it appears the tests fails with Jackrabbit; for now I just made it Oak-only. Please review.

reschke:
The test fails with Jackrabbit because the move of the child node (upon stash restore) fails early; moxing the restoration of properties up seems to fix this.

I would propose to treat this as separate TODO.

vault-core/src/test/resources/test-packages/stashing.zip/META-INF/vault/config.xml
```diff
@@ -0,0 +1,86 @@
+<!--
```
kwin, line 1: no need to add that

vault-core/src/test/resources/test-packages/stashing.zip/META-INF/vault/definition/.content.xml
```diff
@@ -0,0 +1,25 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
kwin, line 1: no need to add that

vault-core/src/test/resources/test-packages/stashing2.zip/META-INF/vault/config.xml
```diff
@@ -0,0 +1,86 @@
+<!--
```
kwin, line 1: not necessary, please remove

vault-core/src/test/resources/test-packages/stashing2.zip/META-INF/vault/definition/.content.xml
```diff
@@ -0,0 +1,25 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
kwin, line 1: please remove

kwin (changes requested):
LGTM, overall, just remove the redundant files from the exploded testing packages

reschke:
Will do.

Other ideas:

- move both test packages under common rolder
- change test NS URI to something we could use for all tests (under "https://jackrabbit.apache.org/filevault/"?)

kwin (approved):

PR #277, opened 2023-02-17
JCRVLT-689: bad description of "isStrict" flag

reschke:


vault-core/src/main/java/org/apache/jackrabbit/vault/packaging/impl/PackagingImpl.java
```diff
@@ -101,7 +101,7 @@ public PackagingImpl() {
         @AttributeDefinition(name = "Allowed Users for Package Installation", description = "The authorizable ids or principal names which are allowed to install packages with the 'requireRoot' flag (in addition to 'admin', 'administrators' and 'system'")
         String[] authIdsForRootInstallation();
 
-        @AttributeDefinition(name = "Default Strict Mode", description = "The default value for strict imports. In strict mode every import failure leads to an exception, while with strict mode those failures are just logged but the import continues in a best effort manner")
+        @AttributeDefinition(name = "Default Strict Mode", description = "The default value for strict imports. In strict mode every import failure leads to an exception, while without strict mode those failures are just logged but the import continues in a best effort manner")
```
kwin, line 104:
```suggestion
        @AttributeDefinition(name = "Default Strict Mode", description = "The default value for imports. In strict mode every import failure leads to an exception, while without strict mode those failures are just logged but the import continues in a best effort manner")
```

kwin (approved):

PR #278, opened 2023-02-19
JCRVLT-690 Allow to prevent Session.save() and Session.refresh(false) in

kwin:
Importer.run(...)

Clarify javadoc around persisting the session.


reschke (approved):
LGTM.

PR #279, opened 2023-02-24
JCRVLT-688: Add "generator" information to packages generated by filevault

reschke:


reschke:
This is a minimal approach.

Open questions:

- test coverage
- does this need to be set elsewhere?
- do we need an override?

vault-core/src/main/java/org/apache/jackrabbit/vault/fs/config/MetaInf.java
```diff
@@ -63,7 +63,12 @@ public interface MetaInf {
      * Name of the 'created by' property
      */
     String CREATED_BY = "createdBy";
-    
+
+    /**
+     * Name of the 'created' property
```
kwin, line 68: copy&paste typo.

vault-core/src/main/java/org/apache/jackrabbit/vault/fs/io/AbstractExporter.java
```diff
@@ -201,6 +201,8 @@ public void export(VaultFile parent, boolean noClose)
             // update properties
             setProperty(MetaInf.CREATED, Calendar.getInstance());
             setProperty(MetaInf.CREATED_BY, mgr.getUserId());
+            String moduleName = "org.apache.jackrabbit.vault";
+            setProperty(MetaInf.GENERATOR, moduleName + ":" + getVersion(moduleName, this.getClass()));
```
kwin, line 205: Please also add a dedicated getter in `org.apache.jackrabbit.vault.packaging.PackageProperties`

vault-core/src/main/java/org/apache/jackrabbit/vault/fs/io/AbstractExporter.java
```diff
@@ -398,6 +400,28 @@ private static PackageType detectPackageType(WorkspaceFilter filter)  {
...
     }
 
+    /**
+     * Returns the version of an Oak module.
+     *
+     * @param moduleName the name of the module
+     * @param clazz a class of the module
+     * @return the version (or "SNAPSHOT" when unknown)
+     */
+    private String getVersion(String moduleName, Class<?> clazz) {
```
kwin, line 410: Can we have a test for this?

vault-core/src/main/java/org/apache/jackrabbit/vault/fs/io/AbstractExporter.java
```diff
@@ -398,6 +400,28 @@ private static PackageType detectPackageType(WorkspaceFilter filter)  {
...
+        // borrowed from oak-commons
+        String version = "SNAPSHOT"; // fallback
+        try (InputStream stream = clazz
+                .getResourceAsStream("/META-INF/maven/org.apache.jackrabbit.vault/" + moduleName + "/pom.properties")) {
+            if (stream != null) {
+                Properties properties = new Properties();
+                properties.load(stream);
+                return properties.getProperty("version", version);
+            }
+        } catch (IOException ignored) {
```
kwin, line 420: When is this supposed to happen? Why should we ignore that?

kwin (changes requested):

vault-core/src/main/java/org/apache/jackrabbit/vault/fs/io/AbstractExporter.java
reschke: For the `generator` string?

vault-core/src/main/java/org/apache/jackrabbit/vault/fs/io/AbstractExporter.java
reschke: Well, we are reading from a stream.

I think best effort is perfectly ok here (but we could log the exception).

vault-core/src/main/java/org/apache/jackrabbit/vault/fs/io/AbstractExporter.java
kwin: The high-level API (with individual getters) is the preferred way of accessing package properties rather than using the low-level API (accessing individual keys) as 
a) it doesn't require to deal with key literals or constants
b) allows for caching or lookup in the manifest alternatively

vault-core/src/main/java/org/apache/jackrabbit/vault/fs/io/AbstractExporter.java
reschke: I have added minimal test coverage; but I'm not really sure whether we actually want to expose the method for testing only.

Is there a test that generated a full package, and where we can inspect the properties.xml later on?

kwin (approved):
LGTM

kwin:
@reschke Please extend https://github.com/apache/jackrabbit-filevault/blob/master/src/site/markdown/properties.md with the new property.

reschke:
Like that...: https://github.com/apache/jackrabbit-filevault/commit/7d07686bdd05b87fe816f403e229bee18873f9ee ?

vault-core/src/main/java/org/apache/jackrabbit/vault/packaging/PackageProperties.java
```diff
@@ -239,6 +244,12 @@ public interface PackageProperties {
      */
     String getCreatedBy();
 
+    /**
+     * Returns the name and version of the component that generated the package.
+     * @return the generator or {@code null} if n/a
```
kwin, line 249: Can you add since tag?

vault-core/src/main/java/org/apache/jackrabbit/vault/packaging/impl/JcrPackageDefinitionImpl.java
```diff
@@ -646,6 +646,13 @@ public String getCreatedBy() {
         return get(PN_CREATED_BY);
     }
 
+    /**
+     * {@inheritDoc}
+     */
+    public String getGenerator() {
```
kwin, line 652: Can you add `@Override`

reschke: Not convinced. I agree that "@Override" is better, but shouldn't we do this consistently here? Any way, your choice, just let me know.

PR #280, opened 2023-03-03
JCRVLT-693 Escape values of type UNDEFINED as well

kwin:


reschke:
Hm. As far as I understood, a given property never can have the type "UNDEFINED". How does this happen in practice?

kwin:
It can happen when serializing a value which is not based on a JCR property. look at the linked Sling ticket.

PR #281, opened 2023-03-13
JCRVLT-671 Add Validator SPI for OSGi configurations

kwin:


reschke (commented):
I honestly have no opinion here and of course trust you with the change.

PR #283, opened 2023-03-16
JCRVLT-696: add logging for primary node type changes

reschke:


vault-core/src/main/java/org/apache/jackrabbit/vault/fs/impl/io/DocViewImporter.java
```diff
@@ -984,6 +984,7 @@ private void removeReferences(@NotNull Node node) throws ReferentialIntegrityExc
             if (importMode == ImportMode.REPLACE && !"rep:root".equals(primaryType) && wspFilter.includesProperty(PathUtil.append(node.getPath(), JcrConstants.JCR_PRIMARYTYPE))) {
                 if (!node.getPrimaryNodeType().getName().equals(primaryType)) {
                     vs.ensureCheckedOut();
+                    log.trace("Setting primary node type {} for {}", primaryType, node.getPath());
```
kwin, line 987: Maybe log both original and target primary node type.

reschke: good point

kwin (approved):

PR #284, opened 2023-03-22
Refactor JcrPackageRegistry to reduce code duplication for upload

kwin:
Both upload methods now use the same private method internally


kwin:
@stefanseifert Can you have a look?

PR #287, opened 2023-04-05
JCRVLT-702 Enable forbiddenapis in Maven build

kwin:
The detects issues like JCRVLT-702

### DIFF
--- a/parent/pom.xml
+++ b/parent/pom.xml
@@ -59,6 +59,7 @@ Apache Jackrabbit FileVault is a project of the Apache Software Foundation.
         <!-- for embedded dependencies the newest version should be used -->
         <oak.max.version>1.52.0</oak.max.version>
         <slf4j.version>1.7.25</slf4j.version>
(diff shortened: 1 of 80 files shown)


kwin:
@reschke 64 issues to fix. Do you have capacity to take this over?

reschke:
Yes. But not before next week.

reschke (approved):
LGTM

kwin:
This will break the build, it needs fixing in the same PR. Do you want to pick it up?

kwin:
I now refactored all methods relying on the default charset/locale/timezone. Please check again @reschke.

reschke (commented):
I *believe* we don't need this for String.format, maybe except when formatting floating point numbers?

kwin:
It is easier to always set it than to exclude some checks. Some other format specifiers are locale specific as well (integer)

reschke:
Consistency is good, but I also dislike verbosity :-)

PR #288, opened 2023-04-17
JCRVLT-697: count moves of stashed child nodes towards autosave limit

reschke:


vault-core/src/main/java/org/apache/jackrabbit/vault/fs/api/ImportInfo.java
```diff
@@ -51,6 +51,14 @@ public interface ImportInfo {
      */
     void onCreated(String path);
 
+    /**
+     * Marks that the child node at {@code path} was moved due to stashing.
+     * @param path the path
+     */
```
kwin, line 57: Can you add a `@since` tag

kwin (approved):
LGTM except for one minor nitpick

PR #295, opened 2023-05-16
JCRVLT-707: update to Jackrabbit 2.20.10 and Oak 1.52.0

reschke:


parent/pom.xml
```diff
@@ -373,7 +374,7 @@ Bundle-Category: jackrabbit
             <dependency>
                 <groupId>org.apache.jackrabbit</groupId>
                 <artifactId>oak-jackrabbit-api</artifactId>
-                <version>${oak.version}</version>
+                <version>${oak.version.compile}</version>
```
kwin, line 377: Actually we state that we are backwards compatible with Jackrabbit 2.16+ (in https://jackrabbit.apache.org/jcr/downloads.html#vlt) so if we distinguish let us rather only manage that dependency in that version and not depend on the newer Oak one.

reschke: Jackrabbit 2.16 has reached End Of Life.

reschke: Not convinced. Why would we reference an unmaintained version?

kwin: I know that you disagree here, but again, that is the job of the distribution to ship current runtime dependencies. Applications only depend on the minimum API level version which is required (as long as the newest version is backwards compatible with it, i.e. fulfills the package-import version range). No need to for FileVault to no longer support being compatible with unmaintained API versions.

kwin: I am only talking about the OSGi bundles, for VLT we should ship the newest dependencies.

reschke: > I know that you disagree here, but again, that is the job of the distribution to ship current runtime dependencies.

Yes. But. Shouldn't we encourage users of FileVault to sop using unmaintained versions?

kwin: By stop supporting running on older versions? No! By writing that somewhere and referring to the maintained versions? Yes

reschke: Well, I continue to disagree. The audience we're talking about is not reading documentation until something breaks.

kwin: Funny thing is: The runtime version range for used packages of Jackrabbit API is the same for 2.16 and the newest one. So even depending on the newest doesn’t make any difference (no one will notice). Also Oak 1.48 is unmaintained as well!

kwin: Also we deliberately support Java 8 although it is  unsupported these days. Not even Oak does encourage adoption of Java 17 by no longer supporting Java 11 ;-). For me that falls in the same category.

kwin:
Let me first fix the bnd-resolver-maven-plugin check so that runtime requirements not fulfilled by a given distro are detected again. That is severely broken right now...

reschke:
OK, I'd really like to be able to close JCRVLT-700 and make a release of FileVault. What's the minimal change we need here, and can we do that right now?

kwin:
This PR won't have any impact on JCRVLT-700 as the FileVault bundle at runtime imports the package `org.apache.jackrabbit.spi.commons.conversion` without a version range, i.e. every bundle exporting that package in any version will resolve. If the distribution ships with bundle `org.apache.jackrabbit.jackrabbit-spi-commons` in version 2.20.10 or newer it works, otherwise you run into this bug (no matter which JR Maven dependency is used). Since JR2 is not correctly versioned (i.e. http://svn.apache.org/viewvc/jackrabbit/branches/2.20/jackrabbit-spi-commons/src/main/java/org/apache/jackrabbit/spi/commons/conversion/package-info.java?view=markup has not been modified for https://issues.apache.org/jira/browse/JCR-4917) there is no import package version range to enforce 2.20.10 or newer....

kwin:
This has been superseded by #296 which also updates the maximum Oak/JR versions.

PR #297, opened 2023-06-17
JCRVLT-683 expect import of principal-based Authz to obey acHandling rules

adamcin:
Optionally restore principal ACLs after import

Introduced feature flag: `vault.feature.stashPrincipalPolicies`

@kwin this branch has an amended commit that I don't have permission to force-push to the existing upstream PR branch.

The diff of the amended commit is essentially:
1. to add the feature flag support to the Importer class
2. to fork the original PrincipalBasedIT into a parallel PrincipalBasedStashingIT, instead of changing its assertions in-place, so that the before and after behavior can be tested simultaneously.

I defined the feature flag above to work as a system property for setting a global default value, as well as supporting a package property of the same name for switching the behavior on individual packages. This should allow us to test the new behavior in different contexts and different downstream tools without having to expose the flag as part of the filevault java API. I left all of your stashing and policy enhancement implementation unmodified after determining that the feature flag could be evaluated at a single point in the Importer, just prior to the calling the `restorePrincipalAcls` method, and could be leveraged when disabled to add diagnostic messages to the progress tracker log.


kwin:
Superseded by #306

PR #308, opened 2023-07-13
JCRVLT-700 Embed most recent JR2 dependencies in CLI app

kwin:


kwin:
> somewhat unrelated changes to http*)

Those are transitive dependencies of JR2. If the version is not set in depMgmt it is inherited from parent's depMgmt, which downgrades 4.5.14 to 4.5.13.

PR #328, opened 2024-01-10
JCRVLT-737: test case

reschke:


kwin:
Can we add this to `ReferenceableIdentifiersImportIT`?

reschke:
I have added tests for the remaining policies and also added test expectations based on what filevault currently does.

@kwin - will look into moving the tests next.

kwin:
There is ITs for all policies already in the aforementioned class!

reschke:
Ack, will review. Maybe in the end this turns out to be an empty PR :-)

reschke:
I had a look at ReferenceableIdentifiersImportIT, and the "parent node not found" case does not occur in these tests. Would you be ok i just adding all of the new test class to ReferenceableIdentifiersImportIT for now, or should we analyze why the existing tests do not cover this case?

reschke:
@kwin - please have a look

vault-core-it/vault-core-integration-tests/src/main/java/org/apache/jackrabbit/vault/packaging/integration/ReferenceableIdentifiersImportIT.java
```diff
@@ -401,6 +422,138 @@ public void testImportDupPolicyLegacy() throws RepositoryException, IOException,
...
+    public void testInstallPackage_FORCE_REMOVE_CONFLICTING_ID() throws Exception {
+        test(IdConflictPolicy.FORCE_REMOVE_CONFLICTING_ID, null, null, false, false);
+    }
+
+    @Test
+    public void testInstallPackage_LEGACY() throws Exception {
+        test(IdConflictPolicy.LEGACY, RepositoryException.class, IllegalStateException.class, false, false);
+    }
+
+    private void test(IdConflictPolicy policy, Class<?> expectedException, Class<?> expectedRootCause, boolean expectNewId,
```
kwin, line 445: can we rename to `assertSomething` and replace the `Something` by the thing which is actually asserted here? Only actual test methods being executed directly by JUnit should have `test` in the name.

vault-core-it/vault-core-integration-tests/src/main/java/org/apache/jackrabbit/vault/packaging/integration/ReferenceableIdentifiersImportIT.java
```diff
@@ -401,6 +422,138 @@ public void testImportDupPolicyLegacy() throws RepositoryException, IOException,
...
+        inf.setProperties(props);
+
+        opts.setMetaInf(inf);
+        File pkgFile = File.createTempFile("testImportMovedResource", ".zip");
+        try (VaultPackage pkg = packMgr.assemble(admin, opts, pkgFile)) {
+            return pkg.getFile();
+        }
+    }
+
+    private void installContentPackage(File pkgFile, IdConflictPolicy policy)
```
kwin, line 540: This is contained in the parent class `IntegrationTestBase` already.

vault-core-it/vault-core-integration-tests/src/main/java/org/apache/jackrabbit/vault/packaging/integration/ReferenceableIdentifiersImportIT.java
```diff
@@ -401,6 +422,138 @@ public void testImportDupPolicyLegacy() throws RepositoryException, IOException,
...
+        Node asset2 = testRoot.getNode(srcName);
+        String id2 = asset2.getIdentifier();
+        if (expectNewId) {
+            assertNotEquals(id1, id2);
+        } else {
+            assertEquals(id1, id2);
+        }
+    }
+
+    private Node addFileNode(Node parent, String name) throws Exception {
```
kwin, line 506: why not using https://jackrabbit.apache.org/api/2.20/org/apache/jackrabbit/commons/JcrUtils.html#putFile-javax.jcr.Node-java.lang.String-java.lang.String-java.io.InputStream-?

kwin (changes requested):

vault-core-it/vault-core-integration-tests/src/main/java/org/apache/jackrabbit/vault/packaging/integration/ReferenceableIdentifiersImportIT.java
reschke: Done.

vault-core-it/vault-core-integration-tests/src/main/java/org/apache/jackrabbit/vault/packaging/integration/ReferenceableIdentifiersImportIT.java
reschke: Yep.

vault-core-it/vault-core-integration-tests/src/main/java/org/apache/jackrabbit/vault/packaging/integration/ReferenceableIdentifiersImportIT.java
reschke: Do you mean "extractVaultPackage"? That seems to search for the file inside the class path; but in this case it's in a Junit temp folder...

kwin: right, I would still recommend to refactor this method to leverage the high-level VaultPackage.extract(...) instead of using the low-level `Importer.run` only. Also I don't understand why you cannot export with right filter in `exportContentPackage()` and need to overwrite that here.

reschke: I believe the filter statement is not needed here.

reschke: Please see update.

kwin (approved):

PR #330, opened 2024-01-22
repro test for issue OAK-10544; currently fails for Oak but passes for…

reschke:
… Jackrabbit


reschke:
I just verified that when we test with OA 1.62.0, the test passes (it still fails for oak.min.version).

PR #332, opened 2024-02-09
add missing filterSet to scrFilter

Walter-Steiner:
After a lot of searching I found the following error in the class "org.apache.jackrabbit.vault.rcp.impl.RcpTaskImpl":

private static WorkspaceFilter createFilterForExcludes(List<String> excludes) throws ConfigurationException {
    // could be done better
    DefaultWorkspaceFilter srcFilter = new DefaultWorkspaceFilter();
    PathFilterSet filterSet = new PathFilterSet("/");
    for (String path : excludes) {
        filterSet.addExclude(new DefaultPathFilter(path));
    }
    **srcFilter.add(filterSet);**
    return srcFilter;
}

The line marked bold was missing.

### DIFF
--- a/vault-rcp/src/main/java/org/apache/jackrabbit/vault/rcp/impl/RcpTaskImpl.java
+++ b/vault-rcp/src/main/java/org/apache/jackrabbit/vault/rcp/impl/RcpTaskImpl.java
@@ -173,6 +173,7 @@ public class RcpTaskImpl implements Runnable, RcpTask {
         for (String path : excludes) {
             filterSet.addExclude(new DefaultPathFilter(path));
         }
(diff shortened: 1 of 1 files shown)


kwin:
Thanks a lot for the PR @Walter-Steiner. Can you open an according issue in https://issues.apache.org/jira/projects/JCRVLT/summary so that we can track that properly? A test would be nice as well (but this is not mandatory for this to get merged).

Walter-Steiner:
Thanks @kwin for looking into this, I created https://issues.apache.org/jira/browse/JCRVLT-746 for it. Now the commit message does not contain this JIRA ID, but I hope it`s still ok.

PR #333, opened 2024-03-11
JCRVLT-747: support binaryless replication for authorizables and their descendants

reschke:
Added a system property to check binaryless replication below authorizables (currently failing).

To run:
~~~
mvn clean install -Dit.test=BinaryPropertiesIT
~~~

Yes, this will need to be refactored.


vault-core/src/main/java/org/apache/jackrabbit/vault/fs/impl/io/JcrSysViewTransformer.java
```diff
@@ -129,6 +138,22 @@ public List<String> close() throws InvalidSerializedDataException {
...
+            List<DocViewProperty2> props = e.getValue();
+            DocViewSAXHandler.log.trace("on '" + path + "' need to set binary reference properties: " + props);
+
+            try {
+                Node n = parent.getSession().getNode(path);
+                for (DocViewProperty2 prop : props) {
+                    prop.apply(n);
+                }
+            } catch (RepositoryException ex) {
+                throw new InvalidSerializedDataException("cannot apply reference properties to " + path, ex);
```
kwin, line 153: reference property to me sound too much like property of type Reference. I would say:

"Cannot import binary reference properties in " (apply is hard to understand)

vault-core/src/main/java/org/apache/jackrabbit/vault/fs/impl/io/JcrSysViewTransformer.java
```diff
@@ -197,11 +223,32 @@ public void startNode(DocViewNode2 ni) throws RepositoryException {
...
+                        for (Name n : currentSubPath) {
+                            path += "/" + resolver.getJCRName(n);
+                        }
+                        List<DocViewProperty2> skipped = skippedProps.get(path);
+                        if (skipped == null) {
+                            skipped = new ArrayList<>();
+                            skippedProps.put(path, skipped);
+                        }
+                        skipped.add(p);
+                        DocViewSAXHandler.log.trace("On node '{}': '{}' is a reference property (not supported by sysview import)",
```
kwin, line 239:
```suggestion
                        DocViewSAXHandler.log.trace("On node '{}': '{}' is a binary reference property (not supported by sysview import)",
```

PR #334, opened 2024-04-23
JCRVLT-751 - Add a test that demonstrates the issue

tmaret:


vault-core-it/vault-core-integration-tests/src/main/java/org/apache/jackrabbit/vault/packaging/integration/UserContentPackageIT.java
```diff
@@ -105,6 +105,15 @@ public void installUserA_Replace_PkgPath() throws RepositoryException, IOExcepti
         installUserA(ImportMode.REPLACE, true, true);
     }
 
+    @Test
+    public void installUserA_Underscores() throws RepositoryException, IOException, PackageException {
```
reschke, line 109: nit: might be better to unpck the ZIP so we don't have a binary here

kwin: That would also help me understand the actual issue better with just looking at the PR. I am still not sure what node structure (name and type) leads to the issue.

tmaret: Thanks for looking at this @kwin !

I suggest to add a breakpoint [here](https://github.com/apache/jackrabbit-filevault/blob/81fa88f1d7af07e77871e2c0bb03515107f18d74/vault-core/src/main/java/org/apache/jackrabbit/vault/fs/io/Importer.java#L1047) while running the test. You should see that the `info.path` has been transformed to `/home/users/test/6k:test-user-a` while it was `/home/users/test/_6k_test-user-a` in the package.

kwin: To me it seems that this is problem with the export rather than with the import according to https://jackrabbit.apache.org/filevault/vaultfs.html#filename-escaping.

tmaret: Potentially, the actual path we found in the FileVault package when we observed the issue was
```
jcr_root/home/users/_/_6k_LELiIYRQppWQpR3W/
```
which has a `_` parent that may be relevant to reproduce on export.

kwin: According to https://jackrabbit.apache.org/filevault/vaultfs.html#filename-escaping this should be rather `jcr_root/home/users/__6k_LELiIYRQppWQpR3W/`. The question really is how the structure is in the original repo (before export).

kwin: For me this package this IT only proofs that the escaped filename `_6k_LELiIYRQppWQpR3W` is correctly deescaped into `6k:LELiIYRQppWQpR3W`, please try to come up with a test case which shows inconsistencies with a round trip test.

tmaret: Indeed, I could reproduce with the export. Added a test that covers that part.

vault-core-it/vault-core-integration-tests/src/main/java/org/apache/jackrabbit/vault/packaging/integration/UserExportIT.java
```diff
@@ -0,0 +1,148 @@
...
+        props.setProperty(VaultPackage.NAME_GROUP, "jackrabbit/sync");
+        props.setProperty(VaultPackage.NAME_NAME, randomUUID().toString());
+        props.setProperty(VaultPackage.NAME_VERSION, "0.0.1");
+        props.setProperty(PackageProperties.NAME_USE_BINARY_REFERENCES, String.valueOf(true));
+        inf.setProperties(props);
+
+        ExportOptions opts = new ExportOptions();
+        opts.setMetaInf(inf);
+
+        opts.setRootPath(authorizablePath);
```
kwin, line 137: it works once you comment this and the following line. If you really need those options they should be in platform format (not in repo format I guess).

kwin:
Superseded by #336

vault-core-it/vault-core-integration-tests/src/main/java/org/apache/jackrabbit/vault/packaging/integration/UserExportIT.java
kwin: I proposed a fix in together with this test in #336

PR #339, opened 2024-07-12
JCRVLT-761 add method to generate qualified name from SPI name object

kwin:
This allows validator implementations to emit more user-friendly node/property names (leveraging qualified instead of expanded name).


reschke (approved):
LGTM.

PR #342, opened 2024-07-25
JCRVLT-762: Log/report node identifiers that cause import to switch to stashing …

reschke:
…and sysview import, also report on slow stashing progress

### DIFF
--- a/vault-core/src/main/java/org/apache/jackrabbit/vault/fs/impl/io/DocViewImporter.java
+++ b/vault-core/src/main/java/org/apache/jackrabbit/vault/fs/impl/io/DocViewImporter.java
@@ -993,7 +993,7 @@ public class DocViewImporter implements DocViewParserHandler {
         if (identifier.isPresent() && !node.getIdentifier().equals(identifier.get()) && !"rep:root".equals(ni.getPrimaryType().orElse(""))) {
             long startTime = System.currentTimeMillis();
             String previousIdentifier = node.getIdentifier();
-            log.info("Node stashing for {} starting, existing identifier: {}, new identifier: {}, import mode: {}",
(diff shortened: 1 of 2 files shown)


vault-core/src/main/java/org/apache/jackrabbit/vault/fs/impl/io/DocViewImporter.java
```diff
@@ -990,13 +991,20 @@ private void removeReferences(@NotNull Node node) throws ReferentialIntegrityExc
         Optional<String> identifier = ni.getIdentifier();
         // try to set uuid via sysview import if it differs from existing one
         if (identifier.isPresent() && !node.getIdentifier().equals(identifier.get()) && !"rep:root".equals(ni.getPrimaryType().orElse(""))) {
+            long startTime = System.currentTimeMillis();
+            String previousIdentifier = node.getIdentifier();
+            log.info("Node stashing for {} starting, existing identifier: {}, new identifier: {}, import mode: {}",
```
kwin, line 996: With info level that is too verbose. Can we lower to debug?

kwin (changes requested):

vault-core/src/main/java/org/apache/jackrabbit/vault/fs/impl/io/DocViewImporter.java
reschke: Yes.

reschke: FWIW, there's also

---
        for (String referencePath : removableReferencePaths) {
            log.info("Remove reference towards {} at {}", node.getIdentifier(), referencePath);
            session.removeItem(referencePath);
        }
---

kwin (approved):

PR #348, opened 2024-08-08
JCRVLT-767: vlt: potential incorrect identifier comparison

reschke:
(work in progress)

This appears to work, except in Jackrabbit, which is unhappy about the sequence "setProperty( ... uid ...)" followed by "addMixing(mix:ref)".

### DIFF
--- a/vault-core/src/main/java/org/apache/jackrabbit/vault/fs/impl/io/DocViewImporter.java
+++ b/vault-core/src/main/java/org/apache/jackrabbit/vault/fs/impl/io/DocViewImporter.java
@@ -989,8 +989,17 @@ public class DocViewImporter implements DocViewParserHandler {
         VersioningState vs = new VersioningState(stack, node);
         Node updatedNode = null;
         Optional<String> identifier = ni.getIdentifier();
+
+        String nodeUUID = null;
+        try {
+            nodeUUID = node.getUUID();
+        } catch (RepositoryException ex) {
+            // node not referenceable
+        }
+
         // try to set uuid via sysview import if it differs from existing one
-        if (identifier.isPresent() && !node.getIdentifier().equals(identifier.get()) && !"rep:root".equals(ni.getPrimaryType().orElse(""))) {
+        if (identifier.isPresent() && nodeUUID != null && !nodeUUID.equals(identifier.get())
+                && !"rep:root".equals(ni.getPrimaryType().orElse(""))) {
             long startTime = System.currentTimeMillis();
             String previousIdentifier = node.getIdentifier();
             log.debug("Node stashing for {} starting, existing identifier: {}, new identifier: {}, import mode: {}",
@@ -1047,8 +1056,38 @@ public class DocViewImporter implements DocViewParserHandler {
                 }
                 // add remaining mixins (for all import modes)
                 for (String mixin : newMixins) {
+                    Property uuidProp = null;
                     vs.ensureCheckedOut();
-                    node.addMixin(mixin);
+
+                    if (mixin.equals("mix:referenceable") && identifier.isPresent()) {
+                        // if this is mix:ref and ni specifies an identifier,
+                        // try to set it
+                        try {
+                            uuidProp = node.setProperty(Property.JCR_UUID, identifier.get());
+                        } catch (RepositoryException ex) {
+                            log.debug("tried to set jcr:uuid to {} before adding mix:referenceable failed, continuing", ex);
+                        }
+
+                        try {
+                            node.addMixin(mixin);
+                        } catch (RepositoryException ex) {
+                            // try to add the mixin; this fails with classic
+                            // jackrabbit, so undo the change for jcr:uuid and retry
+                            if (uuidProp != null) {
+                                log.debug(
+                                        "failed to add mix:referenceable after setting jcr:uuid, retrying without (uuid will be lost)",
+                                        ex);
+                                // retry once after removing jcr:uuid
+                                uuidProp.remove();
+                                node.addMixin(mixin);
+                            } else {
+                                throw ex;
+                            }
+                        }
+                    } else {
+                        node.addMixin(mixin);
+                    }
+
                     updatedNode = node;
                 }
             }


vault-core/src/main/java/org/apache/jackrabbit/vault/fs/impl/io/DocViewImporter.java
```diff
@@ -989,8 +989,17 @@ private void removeReferences(@NotNull Node node) throws ReferentialIntegrityExc
         VersioningState vs = new VersioningState(stack, node);
         Node updatedNode = null;
         Optional<String> identifier = ni.getIdentifier();
+
+        String nodeUUID = null;
+        try {
+            nodeUUID = node.getUUID();
```
kwin, line 995: > Deprecated. As of JCR 2.0, [getIdentifier()](https://developer.adobe.com/experience-manager/reference-materials/spec/javax.jcr/javadocs/jcr-2.0/javax/jcr/Node.html#getIdentifier()) should be used instead.

(https://developer.adobe.com/experience-manager/reference-materials/spec/javax.jcr/javadocs/jcr-2.0/javax/jcr/Node.html#getUUID())

I am not understanding how that makes a diference because this is all only relevant if the node supposed to be update/replaced has a uuid and IMHO the return value should be the same for getUUID and getIdentifier in case the node is referenceable right?

vault-core/src/main/java/org/apache/jackrabbit/vault/fs/impl/io/DocViewImporter.java
```diff
@@ -1047,8 +1056,38 @@ private void removeReferences(@NotNull Node node) throws ReferentialIntegrityExc
...
                 for (String mixin : newMixins) {
+                    Property uuidProp = null;
                     vs.ensureCheckedOut();
-                    node.addMixin(mixin);
+
+                    if (mixin.equals("mix:referenceable") && identifier.isPresent()) {
+                        // if this is mix:ref and ni specifies an identifier,
+                        // try to set it
+                        try {
+                            uuidProp = node.setProperty(Property.JCR_UUID, identifier.get());
```
kwin, line 1066: isn't the uuid already set through the sysview import?

vault-core/src/main/java/org/apache/jackrabbit/vault/fs/impl/io/DocViewImporter.java
reschke: Yes, but this is the simplest way to find out whether the node is referenceable. The other things I tried all require a lot of work with checking node types.

FWIW, the deprecation of `getUUID()` IMHO was incorrect, because these two have different semantics for non-referenceable nodes.

vault-core/src/main/java/org/apache/jackrabbit/vault/fs/impl/io/DocViewImporter.java
reschke: I thought we get here in case we skipped that...?

vault-core/src/main/java/org/apache/jackrabbit/vault/fs/impl/io/DocViewImporter.java
kwin: checking if the property jcr:uuid is protected should be pretty straightforward

reschke: That sounds good, I'll give that a try.

reschke: so

hasProperty -> getProperty -> getDefinition -> isProtected

?

vault-core/src/main/java/org/apache/jackrabbit/vault/fs/impl/io/DocViewImporter.java
```diff
@@ -989,8 +989,19 @@ private void removeReferences(@NotNull Node node) throws ReferentialIntegrityExc
         VersioningState vs = new VersioningState(stack, node);
         Node updatedNode = null;
         Optional<String> identifier = ni.getIdentifier();
+
+        boolean isUuidProtected = false;
+        if (node.hasProperty(Property.JCR_UUID)) {
+            isUuidProtected = node.getProperty(Property.JCR_UUID).getDefinition().isProtected();
+        } else if (! session.getRepository().getDescriptor(Repository.REP_NAME_DESC).contains("Oak")) {
```
kwin, line 996: why else if? Shouldn't this also overwrite isUuidProtected from line 995 potentially?

reschke: good question; tests passed with Jackrabbit; will have another look

reschke: mifgt have to do with https://jackrabbit.apache.org/oak/docs/differences.html#Identifiers - but code is unclear and needs fixing

reschke: Should better wirh https://github.com/apache/jackrabbit-filevault/pull/348/commits/223bb00051fb67b41e2a01dd2d5f70a7f5ca7a2b

vault-core/src/main/java/org/apache/jackrabbit/vault/fs/impl/io/DocViewImporter.java
```diff
@@ -989,8 +989,19 @@ private void removeReferences(@NotNull Node node) throws ReferentialIntegrityExc
         VersioningState vs = new VersioningState(stack, node);
         Node updatedNode = null;
         Optional<String> identifier = ni.getIdentifier();
+
+        boolean isUuidProtected = false;
+        if (! session.getRepository().getDescriptor(Repository.REP_NAME_DESC).contains("Oak")) {
+            // workaround for non-Oak repos that may not be able to add
+            // mixin:referencable *after* setting jcr:uuid
+            isUuidProtected = true;
```
kwin, line 997: The original idea was to encapsulate impl specific differences in https://github.com/apache/jackrabbit-filevault/blob/master/vault-core/src/main/java/org/apache/jackrabbit/vault/fs/spi/ServiceProvider.java.

reschke: Thanks for the pointer; let's do more research on the "setting of jcr:uuid" first, and the outcome of that would inform what information we would want to add there.

PR #362, opened 2025-03-11
JCRVLT-789: AggregateImpl - avoid iterating over sibling nodes

reschke:


vault-core/src/main/java/org/apache/jackrabbit/vault/fs/api/WorkspaceFilter.java
```diff
@@ -102,6 +103,19 @@ public interface WorkspaceFilter extends Dumpable {
...
+    /**
+     * Matches the given path with all filter roots. For each, if it is an ancestor,
+     * add the name of the first path segment of the remaining filter root "below" path
+     * to the result set.
+     *
+     * @param path Path to check
+     * @return first path segments of non-matched paths, or {@code null} when result set
+     * cam not be computed.
+     */
+    default @Nullable Set<String> getFirstChildNamesOfRootsBelowPath(@NotNull String path) {
```
kwin, line 115: maybe `getDirectChildNamesTowardFilterRoots`?

vault-core/src/main/java/org/apache/jackrabbit/vault/fs/api/FilterSet.java
```diff
@@ -239,12 +243,45 @@ public boolean covers(@NotNull String path) {
...
+    }
+
+    /**
+     * Matches the given path with this filter's root. If it is an ancestor, returns the name of the first
+     * path segment of the remaining filter root "below" path.
+     *
+     * @param path Path to check
+     * @return first path segment of non-matched path, or {@code null} when path not ancestor
+     */
+    public @Nullable String getChildNameBelowFilterRoot(@NotNull String path) {
```
kwin, line 263: likewise `getDirectChildNameTowardsFilterRoot`

vault-core/src/main/java/org/apache/jackrabbit/vault/fs/api/WorkspaceFilter.java
```diff
@@ -102,6 +103,19 @@ public interface WorkspaceFilter extends Dumpable {
      */
     boolean isAncestor(@NotNull String path);
 
+    /**
+     * Matches the given path with all filter roots. For each, if it is an ancestor,
+     * add the name of the first path segment of the remaining filter root "below" path
+     * to the result set.
+     *
+     * @param path Path to check
+     * @return first path segments of non-matched paths, or {@code null} when result set
```
kwin, line 112: The condition when null is returned is not really clear. What if only some roots are no offsprings of the given path?

reschke: It would return null, because in that case we can't optimize? So clarify the docs?

reschke: Attempted to clariffy, please check.

(Note that there may be room for improvements here, but  for now the change is sufficient for my use case)

vault-core/src/main/java/org/apache/jackrabbit/vault/fs/api/FilterSet.java
```diff
@@ -239,12 +243,45 @@ public boolean covers(@NotNull String path) {
     }
 
     /**
-     * Checks if the given item is an ancestor of the root node.
+     * Checks if the given item is an ancestor of the filter's root path.
```
mbaedke, line 246: s/given item/given path

reschke: it's consistent with the method above; should I change both?

kwin: I think what was meant was rather 
`Checks if the given path is an ancestor of the filter's root path`. A (string) path is given, not a JCR item!

vault-core/src/main/java/org/apache/jackrabbit/vault/fs/api/FilterSet.java
```diff
@@ -239,12 +243,45 @@ public boolean covers(@NotNull String path) {
...
+            // truncate before first "/"
+            int slashPos = rel.indexOf('/');
+            if (slashPos >= 0) {
+                rel = rel.substring(0, slashPos);
+            }
+
+            result = rel.isEmpty() ? null : rel;
+        }
+
+        log.debug("getChildNameBelowRoot(root={}, path={}) -> {}", rootMatch, pathMatch, result);
```
kwin, line 282: This is inconsistent with the current name of the method.

vault-core/src/main/java/org/apache/jackrabbit/vault/fs/api/WorkspaceFilter.java
kwin: For me there is three conditions:

1. all filter roots are below given path
2. a filter root is not related (i.e. neither ancestor nor descendent)
3. a filter root is an ancestor of this path

Are you sure that both 2. and 3. should lead to the same return value `null`?

vault-core-it/vault-core-integration-tests/src/main/java/org/apache/jackrabbit/vault/packaging/integration/ManySiblingsIT.java
```diff
@@ -0,0 +1,104 @@
...
+public class ManySiblingsIT extends IntegrationTestBase {
+
+    @BeforeClass
+    public static void initRepository() throws RepositoryException, IOException {
+        initRepository(true, false); // always use BlobStore with Oak
+    }
+
+    // JCRVLT-789
+    @Test
+    public void testManySiblings() throws RepositoryException, IOException {
```
kwin, line 52: I think some more tests with filters above or outside the path with the many children should be added (or this one being extended).

vault-core/src/main/java/org/apache/jackrabbit/vault/fs/api/WorkspaceFilter.java
reschke: Well - `null` is the safe approach, because it means doing exactly what we did before.

And no, I'm not sure. I'm trying to improve one specific use case (for which I happen to have an integration test to verify). We probably can improve more cases, but my personal preference would be to do that when we figure out that it's needed - for instance when the current improvement can't be applied because there is a second completely unrelated filter.

Does this make sense?

vault-core/src/main/java/org/apache/jackrabbit/vault/fs/api/FilterSet.java
reschke: Yes. What I meant was: this is consistent with the Javadoc for other methods here. Should I use the opportunity to fix those as well?

vault-core/src/main/java/org/apache/jackrabbit/vault/fs/api/FilterSet.java
reschke: ack

vault-core/src/main/java/org/apache/jackrabbit/vault/fs/api/FilterSet.java
```diff
@@ -239,12 +243,45 @@ public boolean covers(@NotNull String path) {
...
+     * @param path Path to check
+     * @return first path segment of non-matched path, or {@code null} when path not ancestor
+     */
+    public @Nullable String getDirectChildNameTowardsFilterRoot(@NotNull String path) {
+        String result = null;
+
+        String rootMatch = appendSlashIfNeeded(root);
+        String pathMatch = appendSlashIfNeeded(path);
+
+        if (rootMatch.startsWith(pathMatch)) {
```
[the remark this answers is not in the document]
reschke: Well, we can avoid those for the comparison (which internally normalizes), but we'd need it anyway for the string operations below.

It would be nice to have something like "getDescendantRelativeTo", but right now it's not there (could be inspired by nio Path https://docs.oracle.com/javase/8/docs/api/java/nio/file/Path.html#relativize-java.nio.file.Path-.

vault-core-it/vault-core-integration-tests/src/main/java/org/apache/jackrabbit/vault/packaging/integration/ManySiblingsIT.java
reschke: I have now added a test for case 2, as expected the outcome is unnessary iteration.

reschke:
If have now refactored tests, added coverage for more combinations, and added a case "we know no children are covered". I *believe* the "null" case is gone now, but would like to get feedback before I kill it.

@cc @kwin and @mbaedke

vault-core-it/vault-core-integration-tests/src/main/java/org/apache/jackrabbit/vault/packaging/integration/SiblingIterationIT.java
```diff
@@ -0,0 +1,149 @@
...
+import java.io.IOException;
+import java.util.Properties;
+
+import static org.junit.Assert.assertFalse;
+import static org.junit.Assert.assertNotNull;
+import static org.junit.Assert.assertNull;
+import static org.junit.Assert.assertTrue;
+
+/**
+ * Tests checking whether siblings
```
kwin, line 45: this sentence should have a verb :-)

vault-core/src/main/java/org/apache/jackrabbit/vault/fs/api/FilterSet.java
```diff
@@ -239,12 +243,68 @@ public boolean covers(@NotNull String path) {
...
+
+    /**
+     * Matches the given path with this filter's root. If it is an ancestor, returns the name of the first
+     * path segment of the remaining filter root "below" path. If it's unrelated, return an empty string
+     * (indicating that no child node will ever math). Otherwise return {@code null).}
+     *
+     * @param path Path to check
+     * @return first path segment of non-matched path, or {@code null} when path not ancestor
+     */
+    public @Nullable String getDirectChildNameTowardsFilterRoot(@NotNull String path) {
```
kwin, line 264: A Unit test would be nice for this.

reschke: In the unit tests, I assume?

vault-core-it/vault-core-integration-tests/src/main/java/org/apache/jackrabbit/vault/packaging/integration/SiblingIterationIT.java
reschke: Indeed.

vault-core/src/main/java/org/apache/jackrabbit/vault/fs/api/FilterSet.java
reschke: See https://github.com/apache/jackrabbit-filevault/pull/362/commits/788f046f872005736806c458a43954edc18cd597

kwin:
Looks good now, but ASF Jenkins crashes when executing ITs. Now sure it is related to the one you added.

reschke:
master seemt to have a similar issue: https://ci-builds.apache.org/job/Jackrabbit/job/filevault/job/master/369/cloudbees-pipeline-explorer/

kwin:
Hopefully fixed by https://github.com/apache/jackrabbit-filevault/commit/a5a62070cc93151bf12d08d579bad2cf4c474b00.

kwin (approved):

PR #365, opened 2025-03-24
JCRVLT-796: Update commons-io dependency to 2.17.0

reschke:


parent/pom.xml
```diff
@@ -59,7 +59,7 @@ Apache Jackrabbit FileVault is a project of the Apache Software Foundation.
         <!-- for embedded dependencies the newest version should be used -->
         <oak.max.version>1.52.0</oak.max.version>
         <slf4j.version>1.7.25</slf4j.version>
-        <commons-io.version>2.7</commons-io.version>
+        <commons-io.version>2.17.0</commons-io.version>
```
kwin, line 62: Let us only raise the version we embed not the one we depend on.

vault-cli/pom.xml
```diff
@@ -171,7 +171,6 @@
             <dependency>
                 <groupId>commons-io</groupId>
                 <artifactId>commons-io</artifactId>
-                <version>2.11.0</version>
```
kwin, line 174: let's us keep this decoupled from the version we depend on at runtime (within OSGi)

kwin (approved):

kwin:
Build fails due to non-related issue https://issues.apache.org/jira/browse/INFRA-26683.

PR #371, opened 2025-07-01
JCRVLT-748 Fix namespace declaration in XSD

kwin:


kwin:
@reschke I think it is rather best practice to bind a schema to a namespace which is not the null namespace. However I couldn't get the validation to work with that if the filter.xml doesn't define that as default namespace. Do you know a way to overwrite the default namespace in JAXP (https://github.com/apache/jackrabbit-filevault/blob/33917e2119f9166cd4ebcce9ea06abeac513a986/vault-validation/src/main/java/org/apache/jackrabbit/vault/validation/spi/impl/AdvancedFilterValidatorFactory.java#L75)?

PR #378, opened 2025-07-23
JCRVLT-810 skip filter validation

joerghoh:
When controlling both package creation, transport and import checking the filters in import can be time-consuming activity. For that reason it should be possible to skip the validation of the filters, as it will always match. In my test case this reduced the package import time by about 25%.

Right now it only skips the validation of properties, but we could think about skipping the validation of node paths as well. Given that in my case we had around 10 properties per node in average, skipping the filters for adding nodes will just add some small benefit on top.

### DIFF
--- a/vault-core/src/main/java/org/apache/jackrabbit/vault/fs/io/ImportOptions.java
+++ b/vault-core/src/main/java/org/apache/jackrabbit/vault/fs/io/ImportOptions.java
@@ -486,7 +486,7 @@ public class ImportOptions {
 
     /**
      * Allows to skip filter checks when importing packages; you can set this
-     * to {{code true}} if you are sure that there this package contains no content
+     * to {{code true}} if you are sure that this package contains no content
      * which would be rejected by filters.
      * This only affects checks against items in the package, but not existing items in
      * the repository.


joerghoh:
I moved the skip logic out of the workspace filter class, and into the DocViewImporter.

vault-core/src/main/java/org/apache/jackrabbit/vault/fs/config/DefaultWorkspaceFilter.java
```diff
@@ -241,6 +241,10 @@ public void setSkipFilterChecksOnImport(boolean skipFilterChecksOnImport) {
         this.skipFilterChecksOnImport = skipFilterChecksOnImport;
     }
 
+    public boolean getSkipFilterChecksOnImport() {
```
kwin, line 244: If this only affects import I would rather move it to `ImportOptions` and also clarify that this only affects checks against package items not about existing repo items.

joerghoh: ``ImportOptions`` already has such a method; i clarified the javadoc as recommended

vault-core-it/vault-core-integration-tests/src/main/resources/test-packages/tmp-content-outside-filters.zip
kwin, line 1: Let's use non binary packages (i.e. exploded ones, as can be seen e.g. in https://github.com/apache/jackrabbit-filevault/tree/master/vault-core-it/vault-core-integration-tests/src/main/resources/test-packages/import_modes_test_filehandler.zip)

vault-core/src/main/java/org/apache/jackrabbit/vault/fs/config/DefaultWorkspaceFilter.java
kwin: No need to add new methods/fields to WorkspaceFilter then, just use the one from ImportOptions directly (i.e. pass as boolean argument to `DocViewImporter`

kwin (changes requested):

kwin (changes requested):

vault-core/src/main/java/org/apache/jackrabbit/vault/fs/config/DefaultWorkspaceFilter.java
joerghoh: I went through and changed them ... a bit less changes than I actually anticipated.

vault-core/src/main/java/org/apache/jackrabbit/vault/fs/io/ImportOptions.java
```diff
@@ -480,6 +484,26 @@ public void setIdConflictPolicy(@NotNull IdConflictPolicy idConflictPolicy) {
         this.idConflictPolicy = idConflictPolicy;
     }
 
+    /**
+     * Allows to skip filter checks when importing packages; you can set this
+     * to {{code true}} if you are sure that there this package contains no content
```
kwin, line 489:
```suggestion
     * to {{code true}} if you are sure that this package contains no content
```

vault-core/src/main/java/org/apache/jackrabbit/vault/fs/io/ImportOptions.java
```diff
@@ -486,7 +486,7 @@ public void setIdConflictPolicy(@NotNull IdConflictPolicy idConflictPolicy) {
 
     /**
      * Allows to skip filter checks when importing packages; you can set this
-     * to {{code true}} if you are sure that there this package contains no content
+     * to {{code true}} if you are sure that this package contains no content
      * which would be rejected by filters.
      * This only affects checks against items in the package, but not existing items in
```
kwin, line 491:
```suggestion
     * This only affects items in the package, but not existing items in
```

vault-core-it/vault-core-integration-tests/src/main/java/org/apache/jackrabbit/vault/packaging/integration/ImportIT.java
```diff
@@ -445,4 +445,55 @@ public void testEnhancedFileAggregatePackageWithIntermediateSaves() throws IOExc
...
+    public void testSkipFilterChecksOnImport_disabled() throws Exception {
+        ImportOptions opts = getDefaultOptions();
+        Importer importer = new Importer(opts);
+        try (Archive archive = getFileArchive("/test-packages/outside-filters.zip")) {
+            archive.open(true);
+            importer.run(archive, admin.getRootNode());
+            admin.save();
+        }
+        assertNodeExists("/tmp/foo");
+        assertPropertyMissing("/tmp/foo/bar/tobi/excludeddProp");
```
kwin, line 459: Please add some assertions what happens to nodes below filter roots not contained in the package.

PR #409, opened 2026-01-20
JCRVLT-836: log durations and number of nodes in AggregateImpl node walk

reschke:
Summary:

Why `nanoTime` instead of `currentTimeMillis`:

- the latter is wall clock and thus not optimal for measuring elapsed time; furthermode granularity of `currentTimeMillis` is platform dependant and can be > 1ms
- isn't `nanoTime` slower: yes, but these are inbetween IO
- why `Duration` as well? -> can do < 1ms, more readable for big numbers of ms

### DIFF
--- a/vault-core/src/main/java/org/apache/jackrabbit/vault/fs/impl/AggregateImpl.java
+++ b/vault-core/src/main/java/org/apache/jackrabbit/vault/fs/impl/AggregateImpl.java
@@ -29,6 +29,7 @@ import javax.jcr.Value;
 
 import java.io.IOException;
 import java.lang.ref.WeakReference;
+import java.time.Duration;
 import java.util.ArrayList;
 import java.util.Collection;
 import java.util.Collections;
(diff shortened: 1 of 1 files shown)


vault-core/src/main/java/org/apache/jackrabbit/vault/fs/impl/AggregateImpl.java
```diff
@@ -719,10 +737,15 @@ private void prepare(Node node, boolean descend) throws RepositoryException {
         // get a node iterator suitable for the current node and the applicable filters
         NodeIterator nIter = getNodeIteratorFor(node, filter);
 
+        log.debug("starting aggregate walk of '{}'", this.path);
+        int visited = 0;
+        long startTime = System.nanoTime();
```
kwin: Why measure in nanos when emitting in ms only?

reschke:
Of course we can use `Stopwatch` from commons-lang, but guess what timer is uses.

kwin:
Why not System.currentTimeMillis() without any additional helpers?

kwin (approved):
Thanks for clarifications.

reschke:
Thanks; I added a summary to the top, and will also avoid getting the nano time if we're not in DEBUG.

PR #416, opened 2026-02-26
JCRVLT-830 check if a subfolder can be removed

joerghoh:
Implemented some additional checks to prevent the removal of unexpected files.


vault-core/src/main/java/org/apache/jackrabbit/vault/fs/api/WorkspaceFilter.java
```diff
@@ -182,4 +182,25 @@ void dumpCoverage(@NotNull Session session, @NotNull ProgressTrackerListener lis
...
+     * When this method returns {@code true}, an importer may safely remove the existing node at
+     * the path and replace it and its children with the package content. When it returns {@code false},
+     * removal should be avoided or done selectively.
+     *
+     * @param session the session to use for traversing the subtree
+     * @param path    the node path to check (subtree root, must exist)
+     * @return {@code true} if every node and property in the subtree is included and mode is REPLACE
+     * @throws RepositoryException if the path does not exist or traversal fails
+     */
+    boolean isSubtreeFullyOverwritten(@NotNull Session session, @NotNull String path) throws RepositoryException;
```
kwin:
```suggestion
    boolean isSubtreeFullyCovered(@NotNull Session session, @NotNull String path) throws RepositoryException;
```
This is more in line with existing method names

vault-core/src/main/java/org/apache/jackrabbit/vault/fs/api/WorkspaceFilter.java
```diff
@@ -182,4 +182,25 @@ void dumpCoverage(@NotNull Session session, @NotNull ProgressTrackerListener lis
      * @return {@code true} if the property is included in the filter
      */
     boolean includesProperty(String propertyPath);
+
+    /**
+     * Returns whether the given path's subtree is supposed to be fully overwritten during import,
+     * by traversing the repository and checking that every node and property in the subtree is
+     * included by this filter. Returns {@code true} only when:
+     * <ul>
+     *   <li>the path is covered by this filter and import mode is {@link ImportMode#REPLACE},</li>
```
kwin, line 191: I would leave out the first two conditions from here, they seem orthogonal. That way you don't need to pass in the Session.

kwin (changes requested):

vault-core/src/main/java/org/apache/jackrabbit/vault/fs/api/WorkspaceFilter.java
joerghoh: Yes, I can just pass the node (and resolve the path from it), so we have just a single parameter. But even then we can still have these 2 conditions in there.

kwin: For testing purposes I would separate those two aspects: One about the current repo and the other one only taking into account the workspace filter.

joerghoh: I now pass a node.

vault-core/src/main/java/org/apache/jackrabbit/vault/fs/config/DefaultWorkspaceFilter.java
```diff
@@ -278,6 +280,45 @@ public boolean includesProperty(String propertyPath) {
...
+            return false;
+        }
+        PropertyIterator props = node.getProperties();
+        while (props.hasNext()) {
+            Property prop = props.nextProperty();
+            if (!includesProperty(prop.getPath())) {
+                return false;
+            }
+        }
+        NodeIterator children = node.getNodes();
```
kwin, line 336: For performance reason I wouldn't iterate over all properties/children from the repo but just over all filter rules. That is enough to determine the outcome and should be way faster....

joerghoh: ok, so you mean, that if the node is not affected, any childnode of it (and their properties) cannot be affected as well?

kwin: A filter without includes/excludes always covers everything below

vault-core/src/main/java/org/apache/jackrabbit/vault/fs/api/WorkspaceFilter.java
kwin: That was not what I was implying. I would go without Session/Node (and any JCR operations in this class)

vault-core/src/main/java/org/apache/jackrabbit/vault/fs/config/DefaultWorkspaceFilter.java
joerghoh: And what about a filter with includes and excludes? In that case we would need to iterate down the tree...

(given the frequency at which filters without includes and excludes occur, I can definitely add a shortcut without repo access.)

joerghoh:
I made this feature configurable via OSGI. That should be only a temporary measure until we are sure that there are no side-effects. But it required a series of changes to otherwise unrelated classes to pass on that flag.

I added the OSGI property to both the FSPackageRegistry and the JcrPackageRegistry. Default is to have this feature enabled.

joerghoh:
The initial focus for the review should be to make sure that the existing behavior (which can delete content unintentionally) can be preserved with the correct configuration. 

I understand that that adding this configuration increases the size of the change quite a bit, so I wonder if a system property (or env variable) would be a better solution to keep the change smaller.

vault-core/src/main/java/org/apache/jackrabbit/vault/fs/config/DefaultWorkspaceFilter.java
joerghoh: Moving it to an extra ticket JCRVLT-846

reschke:
> The initial focus for the review should be to make sure that the existing behavior (which can delete content unintentionally) can be preserved with the correct configuration.
> 
> I understand that that adding this configuration increases the size of the change quite a bit, so I wonder if a system property (or env variable) would be a better solution to keep the change smaller.

...would it avoid changes in APIs thus package versions?

vault-core/src/main/java/org/apache/jackrabbit/vault/fs/config/DefaultWorkspaceFilter.java
```diff
@@ -278,6 +287,61 @@ public boolean includesProperty(String propertyPath) {
...
+    }
+
+    private boolean isSubtreeFullyOverwrittenRecursive(javax.jcr.Node node) throws RepositoryException {
+        String nodePath = node.getPath();
+        if (!contains(nodePath)) {
+            return false;
+        }
+        PropertyIterator props = node.getProperties();
+        while (props.hasNext()) {
+            Property prop = props.nextProperty();
```
reschke, line 331: Don't we need to check *all* properties here?

I'll assume this is correct, so maybe add a short comment here?

vault-core/src/main/java/org/apache/jackrabbit/vault/fs/config/DefaultWorkspaceFilter.java
```diff
@@ -278,6 +287,61 @@ public boolean includesProperty(String propertyPath) {
...
+        if (getCoveringFilterSet(path) == null) {
+            return false;
+        }
+        if (getImportMode(path) != ImportMode.REPLACE) {
+            return false;
+        }
+        return isSubtreeFullyOverwrittenRecursive(subTree);
+    }
+
+    private boolean isSubtreeFullyOverwrittenRecursive(javax.jcr.Node node) throws RepositoryException {
```
reschke, line 324: as this one is something the first one descends into, should't the method names be consistent?

vault-core/src/main/java/org/apache/jackrabbit/vault/fs/config/DefaultWorkspaceFilter.java
```diff
@@ -278,6 +287,61 @@ public boolean includesProperty(String propertyPath) {
...
+        PropertyIterator props = node.getProperties();
+        while (props.hasNext()) {
+            Property prop = props.nextProperty();
+            if (!includesProperty(prop.getPath())) {
+                return false;
+            }
+        }
+        NodeIterator children = node.getNodes();
+        while (children.hasNext()) {
+            if (!isSubtreeFullyOverwrittenRecursive(children.nextNode())) {
```
reschke, line 338: Hmmm. What about large collections?

vault-core/src/main/java/org/apache/jackrabbit/vault/fs/config/DefaultWorkspaceFilter.java
joerghoh: I don't understand; the code iterates over all properties here.

vault-core/src/main/java/org/apache/jackrabbit/vault/fs/config/DefaultWorkspaceFilter.java
joerghoh: I think that there is no way to avoid iterating all children. But that's a good point, I need to re-evaluate again under which circumstances this code is executed at all.

PR #420, opened 2026-05-21
Explain in more detail how filters are chosen

santiagozky:
giving an example of filters with overlapping paths


kwin (changes requested):
Looks good, can you just add a link towards https://issues.apache.org/jira/browse/JCRVLT-96 (although unlikely to be implemented soon).

santiagozky:
@kwin sure. done.

kwin:
@santiagozky Thanks, looks good now.

PR #427, opened 2026-07-15
Wire the security threat model for agent discoverability (AGENTS.md + SECURITY.md); rename draft-THREAT-MODEL.md -> THREAT_MODEL.md

potiuk:
**This is a proposal for the Jackrabbit PMC to review — please correct, reject, or discuss as needed.** Nothing here is a requirement; the maintainer is the decision-maker.

Thanks for merging the threat-model draft, Angela. Two small follow-ups make it mechanically discoverable to automated security scanners (and to any agent that follows the conventional file layout):

**1. Wire the discoverability chain.** This adds `AGENTS.md` (a `## Security` pointer) and `SECURITY.md`, so the conventional `AGENTS.md → SECURITY.md → THREAT_MODEL.md` chain resolves. An automated scanner mechanically follows that chain to locate the model; without it, the model file is present but the scanner can't find it — discoverability is the one hard gate before a scan can run. *(On jackrabbit-oak this appends a `## Security` pointer to the existing `AGENTS.md` rather than creating a new one.)*

**2. Rename `draft-THREAT-MODEL.md` → `THREAT_MODEL.md` (optional proposal).** Two reasons:
- **It's no longer a draft.** The PMC has reviewed and merged it, so the `draft-` prefix now reads as "not yet adopted" when it is, in fact, the project's adopted model — the name undersells it.
- **`THREAT_MODEL.md` is the conventional, discoverable name.** It's what the `AGENTS.md → SECURITY.md → model` chain and the scanners look for by default, and it matches the layout used across the other Apache repos in this effort (e.g. `apache/solr`). The rename also normalizes the separator (`THREAT-MODEL` → `THREAT_MODEL`).

If you'd rather keep the `draft-` name (or pick a different one), just say so — the rename is the optional part; the wiring is the substantive fix, and we'll point `SECURITY.md` at whatever filename you prefer.

Context: the ASF Security team is preparing the project for an automated agentic security scan. This PR changes **no model content** — only the file name and the discoverability wiring around it.


Amoratinos (approved):
LTGM

kwin:
Just for the record: This broke the build due to the missing ASF License header in https://github.com/apache/jackrabbit-filevault/blob/a2d8ea503cf48aef8b124ec97694eb3a09f9917c/AGENTS.md. @potiuk Any reason not including it (token usage?) What is best practice here? I raised https://github.com/apache/jackrabbit-filevault/pull/429 to add that.

potiuk:
Ah .. Sorry for that. This is really the part that is variable for repository. Yes token usage is a concern.

* We check for rat checks in repo and regular other licence checks - and if the PR fails licence checks, we add the licence
* New RAT 0.18.0 allows to use short version (SPDX) of licence - especially in case those AGENTS.md are also excluded during the source code package preparation. 
* There is an ongoing debate [1] whether short version of licence is acceptable for released source packages (and what form it should be)  - and there are different opinions, but I think there is a consensus that whatever unambigious form of licence is added in non-source-package-released files it is fine.

In this case - no licence was added because the PR did not fail so my automation did not try to see which header to add 😞. 

But I am happy to do a follow up - unless you are happy with what you have now in #429. Just let me know which option is good for you:

a) Full ASF licence
b) SPDX, short licence - non-controversial if you also exclude the AGENTS.md from your release source package

```
<!-- SPDX-License-Identifier: Apache-2.0
https://www.apache.org/licenses/LICENSE-2.0 -->
```

This is the shortest non-ambigious way I found and this is for example what we use in Airflow (and we exclude the AGENTS.md from released sources via .gitattributes https://github.com/apache/airflow/blob/main/.gitattributes

But also - depending on how you do your check, it might also need to add exclusions elsewhere in your pipeline (usually I find it out while raising PR - because PR fails with "missing licence" error :D. 

More details:

[1] https://issues.apache.org/jira/browse/LEGAL-733

PR #428, opened 2026-07-16
JCRVLT-848 adding warn before the node gets dropped as part of the im…

pat-lego:
JCRVLT-848 adding warn before the node gets dropped as part of the import for better tracking of content import process


vault-core/src/main/java/org/apache/jackrabbit/vault/fs/impl/io/FolderArtifactHandler.java
```diff
@@ -169,6 +173,7 @@ public ImportInfoImpl accept(
                                 getAclManagement().clearACL(node);
                             }
                         } else {
+                            log.warn("Replace path: {}", child.getPath());
```
joerghoh: Assume I have a package which removes a subtree with 1000 nodes, how frequently will this message be logged?

vault-core/src/main/java/org/apache/jackrabbit/vault/fs/impl/io/DocViewImporter.java
```diff
@@ -552,6 +552,7 @@ && isIncluded(child, child.getDepth() - rootDepth)) {
                             }
 
                             if (shouldRemoveChild) {
+                                log.warn("Replacing path {}", path);
```
joerghoh: nitpick: I wonder if "replace" is the correct term; in the context of this change I see the term "remove" being used, which is a slightly different thing.

reschke:
Also, do we really need "WARN" here?

kwin:
We do have two different ways of logging during package installation:
1) Server side only via SLF4J
2) Logging exposed to client

Who is supposed to see this message/warn?

joerghoh:
@kwin  I think that the installation log itself contains that information already. But unless this installation is triggered via API and these "installation logs" are also dumped to the (slf4j) log, there is no trace that this happened.

For example if you are using the AEM package manager, these deletes are logged to stdout; if you as a user don't store these logs -- normally no one is doing that -- there is no obvious track that this happened. But in this particular case it might be better to modify the package manager to store a copy of these logs "somewhere", instead of just logging some actions without further context.

vault-core/src/main/java/org/apache/jackrabbit/vault/fs/impl/io/DocViewImporter.java
pat-lego: Fair point I guess I saw it as "Replacing" because it may come back I know for now it is being removed but the filters may have this node in scope and so it could come back.

vault-core/src/main/java/org/apache/jackrabbit/vault/fs/impl/io/FolderArtifactHandler.java
pat-lego: Understandable I would say this could happen yes, but such large packages would also result in other issues such as heavy IO, which would lead me to believe this is not what you are going to worry about in that situation.

pat-lego:
https://github.com/apache/jackrabbit-filevault/pull/428#issuecomment-5001515463 

@reschke I felt that a node delete warrants a WARN especially if you are not anticipating it to be deleted but I am open to ideas just not DEBUG because then we cant do an RCA since the action will have already happened and most of the time the system is not in DEBUG mode.

pat-lego:
@kwin  I agree with what @joerghoh  is saying, often times users build the package (so its not installed) on Author and then replicate that package to the publishers (since most publishers do not offer a package manager). By doing this the content is effectively still on Author instances but missing on Publishers. Users are then confused as to what happened and have no visibility as to what caused this as no information is logged.

Hence this PR aims to add some visibility as to the actions vault does during an installation process.

kwin:
`WARN` level is something which should be actionable. In 99% of the cases this is expected and intended so `WARN` is confusing. IMHO the audit log of the underlying repository gives already everything you would need.

pat-lego:
Ive changed the level to INFO, addressed the verbiage used and I also noticed the file handler was not addressed.

vault-core/src/main/java/org/apache/jackrabbit/vault/fs/impl/io/FolderArtifactHandler.java
joerghoh: I am rather worried about the noise in the logs it creates

pat-lego: Then how can we capture this information in an intelligent way to help with observability because this is not the first occurence of an issue like this and the lack of observability creates issues that are complex to resolve.

reschke:
Logs could be throttled or information be summarized.

kwin:
How is modifying nodes via replication of packages fundamentally different from other ways. Again, why is audit logging not enough?

pat-lego:
Tail the audit.log auditlog.log file while you install a JCR package nothing is printed.

kwin:
I was referring to Oak's built in logging in https://github.com/apache/jackrabbit-oak/blob/c335cb6709c1ba83084a7a16d62fd2981a17fde1/oak-jcr/src/main/java/org/apache/jackrabbit/oak/jcr/delegate/SessionDelegate.java#L730. That logs with DEBUG level by default (so maybe your application (AEM?) needs adjusting log levels), but that should IMHO cover that already. 
In addition there is a dedicated write operation logger (https://issues.apache.org/jira/browse/OAK-2011).
What other information do you need @pat-lego?

pat-lego:
> I was referring to Oak's built in logging in https://github.com/apache/jackrabbit-oak/blob/c335cb6709c1ba83084a7a16d62fd2981a17fde1/oak-jcr/src/main/java/org/apache/jackrabbit/oak/jcr/delegate/SessionDelegate.java#L730. That logs with DEBUG level by default (so maybe your application (AEM?) needs adjusting log levels), but that should IMHO cover that already. In addition there is a dedicated write operation logger (https://issues.apache.org/jira/browse/OAK-2011). What other information do you need @pat-lego?

I see I had it in INFO which is why I was not seeing anything, these logs would be perfect to be honest, the issue is that these logs are not tracked and we never have it on DEBUG. 

The reason why it is not in DEBUG is because what @joerghoh has been referencing, I had time over the weekend to think of a way to implement this without requiring a lot of logging. For now I think we can close this out as I work on the implementation that does not require large volumes of logs to be generated.

